PR #35, opened 2019-09-27
Tech and Content Review - Seismic Sampling Method (Revision)

KellyVance:
Reconciliation to Collections and elimination of non-selectable terms as concepts
Content unchanged
Hierarchy reordered to flat list and collections established to accommodate categorisation.

As per Data Integrity Team (DIT) review and advice (26-09-2019).

### DIFF
--- a/vocabularies/seismic-sampling-method.ttl
+++ b/vocabularies/seismic-sampling-method.ttl
@@ -177,10 +177,5 @@
         :weight-drop ;
     skos:prefLabel "Onshore"@en .
 
-:vibration a skos:Collection ;
-    skos:definition "A mass is held in contact with the ground and oscillated to put a range of frequencies into the earth."@en ;
-    skos:member :noise-source,
-        :vibroseis ;
-    skos:prefLabel "Vibration"@en .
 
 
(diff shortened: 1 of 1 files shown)


KellyVance:
> I presume concept:AWD on line 173 occurring after the collection group makes no tangible difference in the code. Just a bit out of place visually when reading through.
> Liking the breakdown of the sources into their respective collection

@CantRoger 
You are correct that there is no material difference.
But for the sake of best practice and human readability I have moved AWD to the top and re-ordered the collections to group the source types and the environments together.
All already approved content has been unaltered.

CantRoger:
> I presume concept:AWD on line 173 occurring after the collection group makes no tangible difference in the code. Just a bit out of place visually when reading through.
> Liking the breakdown of the sources into their respective collection
> 
> @CantRoger
> You are correct that there is no material difference.
> But for the sake of best practice and human readability I have moved AWD to the top and re-ordered the collections to group the source types and the environments together.
> All already approved content has been unaltered.

Vance, is vibration:collection now duplicated on lines 153 and 180?

PR #37, opened 2019-10-02
Update data-access-rights.ttl - Inclusion of Protected

KellyVance:
Addition of 'Protected access' as a sub-category of 'Restricted access' and as member of 'Closed data access rights' collection


johnkirsten:
All good, thanks for adding this

KellyVance:
> Should this be a narrower definition of `restricted` or a top concept of this vocab?

Changed vocabulary to move Protected Access from narrower of Restricted to a top concept.

Technically I think protected _is_ a subset of restricted. But as a data management function it will probably be easier to deal with as an equally ranked topConcept.

PR #50, opened 2019-10-17

johnmckellar:
Major changes to "non-current" and omission of "non-current-unfinalized", plus changes suggested by Vance.


KellyVance:
@dxwell @DavidCrosswellGSQ 
I am unsure the topConcepts are real concepts. Is this another problem that needs a Collections solution? Thoughts?

PR #51, opened 2019-10-22

johnmckellar:
Following consultation with Jodie, the vocab has been considerably reduced:
1. Top concepts reduced to: Application, Granted, Non-current
2. Anything to do with payment has been accordingly omitted
3. A competing application for a Resource Permit represents an application made on the same day for the same mineral; it thus applies only to EPMs.
4. See definition for conditionally surrendered!!

Please expedite


KellyVance:
@johnmckellar changes can be made directly in the dev branch. GitHub will track the changes and the dev branch is for the development of the vocab. I'll push this pull request through and let the review happen in the dev branch.

PR #53, opened 2019-10-28

johnmckellar:
Adding the new definitions (please review) from Jodie and eliminating: on-hold oil shale; expired under granted; and accepted under granted. We can sort out the remainder tomorrow, 29.10.2019.
Cheers


KellyVance:
Hey John. Need to have a chat about this to match changes to discussions on Friday.

johnmckellar:
I added the definitions as we still need them, and the concepts that were eliminated were to be eliminated in any case.

PR #57, opened 2019-11-05
Content Review - Borehole Design

KellyVance:
GraphDb to VocPrez (with link) and general review @DavidCrosswellGSQ 
Custodian - @mark-gordon-dnrme 
Peer Reviewer - @LizDerrington 

Technical Review has been done by @LukeHauck

### DIFF
--- a/vocabularies/borehole-design.ttl
+++ b/vocabularies/borehole-design.ttl
@@ -24,27 +24,27 @@ bhd:conceptScheme a skos:ConceptScheme ;
 
 bhd:deviated a skos:Concept ;
     skos:altLabel "Directional"@en ;
-    skos:definition "A borehole drilled from its origin point with directional geosteering equipment that intends to change inclination and/or azimuth to a specified trajectory to intersect its target or targets."@en ;
+    skos:definition "A borehole drilled with geosteering equipment with the intent of controlling inclination and/or azimuth on a specified trajectory to intersect a target or targets."@en ;
     skos:inScheme bhd:conceptScheme ;
     skos:prefLabel "Deviated"@en ;
     skos:topConceptOf bhd:conceptScheme .
 
 bhd:inclined a skos:Concept ;
     skos:altLabel "Slant"@en ;
-    skos:definition "A borehole that is drilled from its origin point at a non-vertical angle without specific geosteering equipment, but with the intent of maintaining that trajectory."@en ;
+    skos:definition "A borehole drilled from its origin point at a non-vertical angle without geosteering equipment with the intent of maintaining its initial trajectory."@en ;
     skos:inScheme bhd:conceptScheme ;
     skos:prefLabel "Inclined"@en ;
     skos:topConceptOf bhd:conceptScheme .
 
 bhd:lateral a skos:Concept ;
     skos:altLabel "Horizontal"@en ;
-    skos:definition "A borehole drilled with geosteering equipment to intersect a specific stratigraphic interval with the intent of drilling laterally to maximise exposure to that specific interval."@en ;
+    skos:definition "A borehole drilled with geosteering equipment to intersect a specific stratigraphic interval with the intent of drilling laterally within the interval to maximise exposure."@en ;
     skos:inScheme bhd:conceptScheme ;
     skos:prefLabel "Lateral"@en ;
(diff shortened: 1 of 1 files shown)


vocabularies/borehole-design.ttl
```diff
@@ -0,0 +1,51 @@
...
+    skos:definition "A description of the intended configuration and trajectory of a borehole."@en ;
+    skos:hasTopConcept bhd:deviated,
+        bhd:inclined,
+        bhd:lateral,
+        bhd:vertical ;
+    skos:prefLabel "Borehole Design"@en .
+
+bhd:deviated a skos:Concept ;
+    skos:altLabel "Directional"@en ;
+    skos:definition "A borehole drilled from its origin point with directional geosteering equipment that intends to change inclination and/or azimuth to a specified trajectory to intersect its target or targets."@en ;
```
LizDerrington: Maybe change to:- 'A borehole drilled from its origin point with geosteering equipment with the intent of controlling inclination and/or azimuth on a specified trajectory to intersect a target or targets'.

vocabularies/borehole-design.ttl
```diff
@@ -0,0 +1,51 @@
...
+    skos:inScheme bhd:conceptScheme ;
+    skos:prefLabel "Deviated"@en ;
+    skos:topConceptOf bhd:conceptScheme .
+
+bhd:inclined a skos:Concept ;
+    skos:altLabel "Slant"@en ;
+    skos:definition "A borehole that is drilled from its origin point at a non-vertical angle without specific geosteering equipment, but with the intent of maintaining that trajectory."@en ;
```
LizDerrington: "A borehole that is drilled from its origin point at a non-vertical angle without specific geosteering equipment, and with the intent of maintaining that angled trajectory throughout drilling".

vocabularies/borehole-design.ttl
```diff
@@ -0,0 +1,51 @@
...
+    skos:inScheme bhd:conceptScheme ;
+    skos:prefLabel "Inclined"@en ;
+    skos:topConceptOf bhd:conceptScheme .
+
+bhd:lateral a skos:Concept ;
+    skos:altLabel "Horizontal"@en ;
+    skos:definition "A borehole drilled with geosteering equipment to intersect a specific stratigraphic interval with the intent of drilling laterally to maximise exposure to that specific interval."@en ;
```
LizDerrington: "A borehole drilled with geosteering equipment to intersect a specific targeted interval with the intent of drilling laterally to maximise exposure to that specific interval.

LizDerrington:
The Petroleum and Gas Reporting Guideline (2018) and Submission Template refer to 'Well Design'  which correlates to the title "Borehole Design' as proposed here. 'Well Design' in the Submission Template is defined as 'The intentional design of the well or bore being either vertical or deviated'. Correspondingly the Submission Template (which is PPDM based) offers a drop down menu choice under 'Well Design' of 'Vertical' or 'Deviated'. Using this terminology GA's 'Borehole Inclination' vocab fits in as 'Vertical' = GA Vertical; 'Deviated' = GA horizontal, GA inclined up and GA inclined down.

KellyVance:
This should replace our Borehole Inclination.

We have used the word Design because this describes the intention of the well. It gets messy if you try describing the actual result of the well when you drill a 'vertical' well that naturally drifts during drilling and results in a final inclination of 40° from vertical, which does happen. We also want to avoid describing the final geometry of the well using nominal categories rather than measurements.

With all due respect to GA, CSIRO and EGR, vocabulary they share is very basic and seems to be have designed by an academic asking "what directions could a hole theoretically be drilled?", but without an operational and technical understanding of how wells are drilled and what information is pertinent for their description. At least as far as describing a well in terms industry would understand, that is useful for operational needs, and describes wells in a way that meets the needs and intent of our own regulation I am confident that this Well Design vocabulary is _significantly_ superior for our purposes than adopting from the GA/CSIRO/EGR inclination vocab.

CoalLog and Minerals tend not to describe things by name, they just use the inclination and azimuth pair of the final well, or the full survey if directionally drilled. For PPDM they do not specify the terms that fit into their WELL_DESIGN field, but common industry usage is Vertical and Deviated, with some including lateral or horizontal.

PR #59, opened 2019-11-15
Technical Review - Minerals Vocabulary

KellyVance:
Minerals vocab for technical review

- Please note conceptScheme sourceNote
- Despite some of the parent concepts having collection-sounding names, these are in fact genuine classification groups of mineralogical species. They just don't have catchy names that can be used.
- 6321 concepts, so not feasible to check every one. I suggest a random check of a small number of concepts.
- I have used triple quotation marks in the concepts that i think need them. I can do a bulk update so all definitions use triple quotes if this is needed.


nicholascar:
The definitions which reference other classification and notation systems, such as *Nickel-Strunz Classification: [08.DL.15]* or *NdCu6(AsO4)3(OH)6 · 3H2O* should be represented with `skos:notation` predicates with specialised Datatype Properties, as per the [SKOS Reference, Section 6.5.1](https://www.w3.org/TR/skos-reference/#L2613). For `min:agardite-nd` I would expect to see:

```
min:agardite-nd a skos:Concept ;
    skos:broader min:phosphates-etc-with-additional-anions-with-h2o-with-large-and-medium-sized-cations-oh-etc-ro4-equals-2-1 ;
    skos:definition "SOME TEXT YOU MIGHT LIKE TO GET FROM SOMEWHERE OR MAKE UP"@en ;
    skos:notation "NdCu6(AsO4)3(OH)6 · 3H2O"^^ , "08.DL.15"^^min:nickel-strunz-classification ;
    skos:inScheme min:conceptScheme ;
    skos:prefLabel "Agardite-(Nd)"@en .

min:chemical-formula
    a owl:Class ;
    rdfs:subClassOf rdfs:Literal ;
    skos:prefLabel "Chemical Formula"@en ;
    skos:definition "A chemical formula is a way of presenting information about the chemical proportions of atoms that constitute a particular chemical compound or molecule, using chemical element symbols, numbers, and sometimes also other symbols, such as parentheses, dashes, brackets, commas and plus (+) and minus (−) signs."@en ;
    rdfs:seeAlso <https://en.wikipedia.org/wiki/Chemical_formula> ;
.

min:nickel-strunz-classification
    a owl:Class ;
    rdfs:subClassOf rdfs:Literal ;
    skos:prefLabel "Nickel-Strunz Classification"@en ;
    skos:definition "Nickel–Strunz classification is a scheme for categorizing minerals based upon their chemical composition, introduced by German mineralogist Karl Hugo Strunz."@en ;
    rdfs:seeAlso <https://en.wikipedia.org/wiki/Nickel%E2%80%93Strunz_classification> ;
.
```

I left the '[]' off the NS Classification `skos:notation` value but if they are part of the notation, include them.

KellyVance:
The defining property of a mineral is its chemical composition. So the chemical formula is actually the most appropriate definition. A sentence for each mineral doesn't really make sense, it would be an enormous amount of extra work for something that would be less informative to a mineralogist or geochemist who might use the vocabulary.

The nickel-strunz 'notation' is not a true notation, it is more a classification code. Multiple minerals could be 08.DL.15, so notations would not be unique.

PR #62, opened 2019-11-25
Review readme

KellyVance:
Updated the vocabulary readme to reflect the process as it currently stands. Or at least be much closer to it.


nicholascar:
This has been sitting here for a while but is a simple update and should be merged directly. Can always be re-updated.

PR #84, opened 2020-02-05
Content and Tech Review - Survey Type

KellyVance:
Edition 1 Review
Expedited review to enable SRA development.
For wider distribution and deeper content review for edition 2 after initial development.

Can augment with additional terms as needed after approval.


Greenwoodmatthew:
Base vocab look good.  More Alias and Types to add, but as a functional start good

PR #99, opened 2020-02-17
Please review updated vocabulary - see changes below

DavidCrosswellGSQ:
The following changes were made:
1. Added EDC and MIDCF as notation.
2. Used 'Mount Isa Drill Core Facility' as the preflabel. This will fit better in a drop-down in a system UI.
3. Put 'John Campbell Miles Drill Core Storage Facility' as an altLabel.


KellyVance:
Changes make sense to me.

PR #102, opened 2020-02-17
Update and rename project-class.ttl to resource-project-lifecycle.ttl

DavidCrosswellGSQ:
Changed the name of this vocab to Resource Project Lifecycle


vocabularies/resource-project-lifecycle.ttl
```diff
@@ -19,7 +19,7 @@ prcls:conceptScheme a skos:ConceptScheme ;
     skos:hasTopConcept prcls:appraisal,
         prcls:development,
         prcls:exploration ;
-    skos:prefLabel "Project Class"@en .
```
KellyVance: Should move Project Class to altLabel rather than delete as it is a PPDM published term, i.e.
skos:altLabel "Project Class"@en;

DavidCrosswellGSQ: Alt label added.

PR #109, opened 2020-02-21
Content Review survey relationship type

KellyVance:
Review of controlled list for types of survey relationships.

### DIFF
--- a/vocabularies/survey-relationship-type.ttl
+++ b/vocabularies/survey-relationship-type.ttl
@@ -53,7 +53,8 @@ srt:reprocessing a skos:Concept ;
     skos:topConceptOf srt:conceptScheme .
 
 srt:resampling a skos:Concept ;
-    skos:definition "Generating new physical sample by changing the form of an existing sample."@en ;
+    skos:altLabel "subsampling"@en ;
+    skos:definition "Generating a new physical sample by removing material from, or changing the form of an existing sample."@en ;
     skos:inScheme srt:conceptScheme ;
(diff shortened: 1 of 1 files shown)


vocabularies/survey-relationship-type.ttl
```diff
@@ -0,0 +1,66 @@
...
+    skos:topConceptOf srt:conceptScheme .
+
+srt:reprocessing a skos:Concept ;
+    skos:definition "Generating a new digital sample by reprocessing pre-existing raw data with a new algorithm or processing parameters."@en ;
+    skos:inScheme srt:conceptScheme ;
+    skos:prefLabel "Reprocessing"@en ;
+    skos:topConceptOf srt:conceptScheme .
+
+srt:resampling a skos:Concept ;
+    skos:definition "Generating new physical sample by changing the form of an existing sample."@en ;
```
GSQ-AI: Grammar
> Generating **a** new physical sample

Regarding "changing the form", I feel that this isn't describing the process appropriately.
In my opinion, the process isn't changing the _form_ of the sample, it's removing material from the sample. There is some deformation with this, yes, but the result is an altered original sample and anew sub-sample.

GSQ-AI: @geoderekh

KellyVance: Maybe "Generating new a physical sample from an existing sample." Simplify it, broaden its applicability.

PR #113, opened 2020-02-25
created a basic site-substatus vocab

geoderekh:
vocab content is 1:1 from merlin exploration status. Vocab is needed to cater for mines and minocc having two independent statuses (mine status will be rolled into site-status presently.


KellyVance:
In light of the status v sub-status discussion this vocabulary may need to be deleted and merged with the status vocab. Let's discuss further in person with the necessary persons.

PR #126, opened 2020-03-11
first version - missing concave hull defn

nicholascar:
Hi @KellyVance. I've made this first version of the vocab with all the terms we discussed however I've put Concave & Convex Hull in as narrower terms of Boundary since they are! They are boundaries under certain restrictions: rectangular & circular respectively.

I've not provided a definition for Concave Hull, please could you?


nicholascar:
I've skipped the dev- PR step, as per the updated workflow in README.md, and gone straight to review- since the vocab passes all SHACL validators in shapes/ therefore it's valid RDF and has the (very minimal) metadata and Concept properties we've listed as mandatory.

nicholascar:
Changes made as per @KellyVance suggestions

PR #127, opened 2020-03-13
Updated CRS

DavidCrosswellGSQ:
Changed the Preflabels to the code, e.g. GDA2020.
This is so the drop-down labels show the code, not the long name.

### DIFF
--- a/vocabularies/qld-crs.ttl
+++ b/vocabularies/qld-crs.ttl
@@ -9,90 +9,89 @@
 @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
 
 <http://linked.data.gov.au/def/queensland-crs> a owl:Ontology , skos:ConceptScheme ;
-    dct:created "2020-01-30T15:57:41"^^xsd:dateTime ;
+    dct:created "2020-03-13T15:57:41"^^xsd:dateTime ;
     dct:creator "Geological Survey of Queensland" ;
-    dct:modified "2020-01-30T15:57:41"^^xsd:dateTime ;
+    dct:modified "2020-03-13T15:57:41"^^xsd:dateTime ;
     dct:source "Compiled by the Geological Survey of Queensland" ;
     skos:definition "The coordinate reference systems used and refered to within Queensland"@en ;
-    skos:hasTopConcept qldcrs:australian-geodetic-datum-1966,
-        qldcrs:australian-geodetic-datum-1984,
-        qldcrs:australian-map-grid-1966,
-        qldcrs:australian-map-grid-1984,
-        qldcrs:geocentric-datum-of-australia-1994,
-        qldcrs:geocentric-datum-of-australia-2020,
-        qldcrs:map-grid-of-australia-1994,
-        qldcrs:map-grid-of-australia-2020 ;
+    skos:hasTopConcept qldcrs:AGD66,
+        qldcrs:AGD84,
+        qldcrs:AMG66,
+        qldcrs:AMG84,
+        qldcrs:GDA94,
+        qldcrs:GDA2020,
+        qldcrs:MGA94,
+        qldcrs:MGA2020 ;
     skos:prefLabel "Queensland CRS"@en .
 
 qldcrs:geographic a skos:Collection ;
     skos:definition "A geographic coordinate system (GCS) is a coordinate system which uses a three-dimensional spherical surface (ellipsoid) to define locations on the earth. A common choice of coordinates is latitude and longitude."@en ;
-    skos:member qldcrs:australian-geodetic-datum-1966,
-        qldcrs:australian-geodetic-datum-1984,
-        qldcrs:geocentric-datum-of-australia-1994,
-        qldcrs:geocentric-datum-of-australia-22 ;
+    skos:member qldcrs:AGD66,
+        qldcrs:AGD84,
+        qldcrs:GDA94,
+        qldcrs:GDA2020 ;
     skos:prefLabel "Geographic"@en .
 
 qldcrs:projected a skos:Collection ;
     skos:definition "A projected coordinate system (PCS) is a projection of the geographic coordinate to, on two-dimensional surface (the map). The used to convert between Australian Geodetic Data and their Projected equivalents is the Universal Transverse Mercator projection."@en ;
-    skos:member qldcrs:australian-map-grid-1966,
-        qldcrs:australian-map-grid-1984,
-        qldcrs:map-grid-of-australia-1994,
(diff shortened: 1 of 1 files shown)


vocabularies/qld-crs.ttl
```diff
@@ -9,90 +9,89 @@
...
+        qldcrs:AMG66,
+        qldcrs:AMG84,
+        qldcrs:GDA94,
+        qldcrs:GDA2020,
+        qldcrs:MGA94,
+        qldcrs:MGA2020 ;
     skos:prefLabel "Queensland CRS"@en .
 
 qldcrs:geographic a skos:Collection ;
     skos:definition "A geographic coordinate system (GCS) is a coordinate system which uses a three-dimensional spherical surface (ellipsoid) to define locations on the earth. A common choice of coordinates is latitude and longitude."@en ;
```
geoderekh: A spherical surface is not an ellipsoid. There are both spherical and ellipsoidal geographic CRS so...
Suggested def: A geographic coordinate system (GCS) is a coordinate system which uses a three-dimensional spherical or ellipsoidal surface to define locations on the earth. A common choice of coordinates is latitude and longitude.

vocabularies/qld-crs.ttl
```diff
@@ -9,90 +9,89 @@
...
+        qldcrs:AGD84,
+        qldcrs:AMG66,
+        qldcrs:AMG84,
+        qldcrs:GDA94,
+        qldcrs:GDA2020,
+        qldcrs:MGA94,
+        qldcrs:MGA2020 ;
     skos:prefLabel "Queensland CRS"@en .
 
 qldcrs:geographic a skos:Collection ;
```
geoderekh: Should be named "geographic coordinate systems" 
has prefLabel: "Geographic Coordinate System" @en 
has alfLabel: "Geographic Coordinate Reference System" @en ; "Geographic CRS" @en

vocabularies/qld-crs.ttl
```diff
@@ -9,90 +9,89 @@
...
-        qldcrs:australian-geodetic-datum-1984,
-        qldcrs:geocentric-datum-of-australia-1994,
-        qldcrs:geocentric-datum-of-australia-22 ;
+    skos:member qldcrs:AGD66,
+        qldcrs:AGD84,
+        qldcrs:GDA94,
+        qldcrs:GDA2020 ;
     skos:prefLabel "Geographic"@en .
 
 qldcrs:projected a skos:Collection ;
```
geoderekh: Should be named "projected coordinate systems"
has prefLabel: "Projected Coordinate System" @en
has alfLabel: "Projected Coordinate Reference System" @en ; "Projected CRS" @en

vocabularies/qld-crs.ttl
```diff
@@ -9,90 +9,89 @@
 @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
 
 <http://linked.data.gov.au/def/queensland-crs> a owl:Ontology , skos:ConceptScheme ;
```
geoderekh: There was surely a reason for the change, and I am sure that Nick would be the better person to speak to this, but I suspect there will be problems using notations for class names. 
Presumably we could have the same value for notation and preflabel, or at the least get the lookups populated from the notation in this vocab.

vocabularies/qld-crs.ttl
```diff
@@ -9,90 +9,89 @@
...
     skos:inScheme <http://linked.data.gov.au/def/queensland-crs> ;
     skos:notation "AMG84" ;
-    skos:prefLabel "Australian Map Grid 1984"@en ;
+    skos:prefLabel "AMG84"@en ;
     skos:topConceptOf <http://linked.data.gov.au/def/queensland-crs> .
 
-qldcrs:geocentric-datum-of-australia-1994 a skos:Concept ;
-    skos:definition "The official geodetic datum adopted nationally in June 2020. GDA2020 is a coordinate reference system whose origin coincides with a determination of the centre of mass of the earth, hence the term geocentric. GDA2020 is a plate-fixed or static coordinate datum based on the International Terrestrial Reference Frame 1992 (ITRF92), held at the reference epoch of 1 January 2020."@en ;
+qldcrs:GDA94 a skos:Concept ;
+    skos:definition "Geocentric Datum of Australia 1994. The official geodetic datum adopted nationally in June 2020. GDA2020 is a coordinate reference system whose origin coincides with a determination of the centre of mass of the earth, hence the term geocentric. GDA2020 is a plate-fixed or static coordinate datum based on the International Terrestrial Reference Frame 1992 (ITRF92), held at the reference epoch of 1 January 2020."@en ;
```
geoderekh: The definition and concept for GDA94 and GDA2020 have been mixed up.

vocabularies/qld-crs.ttl
nicholascar: > There was surely a reason for the change

Sorry but looking here I can't see what the change you're referring to is. Do you mean the merger of the Ontology & Concept Scheme (that's where the code line reference is at here).

`skos:notation` should really be used to indicate an ID for a concept from some other system, e.g. the Concept coded with some non-URI in a geology handbook or something. Can have multiple notations for multiple codes. 

I always like to use the primary code for the Concept in the URI if it is well known instead of a tokenised prefLabel but you can have both - the well-known code in the URI and in a `skos:notation`.

nicholascar:
I think an anti-pattern has been introduced here! It's fine - preferable even - to have well-known short codes for the URIs and, if those codes need to be matched for drop-downs and so on, they should be in notation fields (as they are) and perhaps even altLabels for test search matching, but I don't think prefLabels should be based on short codes, even if well known. I do think prefLabels should be the written out titles of the items, so "Geocentric Datum of Australia 2020", rather than "GDA202". This is in keeping with the idea that SKOS has all these facilities for codes and that the prefLabel should be unambiguous fully spelled-out terms, not contractions etc.

For example, if GDA2020 has an "international CRS ID" of ABC that is used for GDA202 in some systems, it would be better that the Concept is called (prefLabel) "Geocentric Datum of Australia 2020" and has notations "GDA202" & "ABC" with the notation datatypes indicating the notation supplier source (the authority).

nicholascar:
As per my previous comment higher up: it is sensible to use DGA202 etc. in the ID part of Concept's URIs though as these are stable and unambiguous and short.

DavidCrosswellGSQ:
Good conversation. I think that the preflabel is still GDA2020 etc. The skos:definition is the full "Geocentric Datum of Australia 2020". People refer to the concept as GDA2020.
The notation could better be the EPSG code (SRID) than a duplicate of the preflabel.

PR #130, opened 2020-03-19

KellyVance:
introduce:
stratigraphic intervals  - for simple strat succession
reservoir intervals - for reservoir intervals and linking to reservoir property data
well completion intervals - to capture contact intervals between the borehole and the target formations


vocabularies/geological-observation-type.ttl
```diff
@@ -46,13 +46,16 @@
         obtyp:proximity-of-orebody-obrackrelated-intervalcbrack,
         obtyp:quartz-textures,
         obtyp:regolith-observations,
+        obtyp:reservoir-interval
```
LukeHauck: needs comma on end.

KellyVance: well spotted. fixed

PR #131, opened 2020-03-19
review and update minsample

geoderekh:
no changes made to the file except moving the list of top concepts to the top. 
Reason for edit and pull request is that we are missing sample_type vocab in master. minsample.ttl looks like it is a skosified version of the MERLIN CPF table equivalent, with merlin codes as altlabels, code names as preflabels. @KellyVance @DavidCrosswellGSQ  unless offline progress has been made with this, lets push this through review and we can add or remove things at a later date.


geoderekh:
my bad, there was one already...

PR #133, opened 2020-03-20
Update resource-project-lifecycle.ttl

KellyVance:
Addition of concepts
Closed
Closed with exploration

As per discussion with Derek to account for all possible permutations of mine and exploration activities, and provide the ability to map from current MinOcc metadata.


vocabularies/resource-project-lifecycle.ttl
```diff
@@ -29,6 +29,21 @@ plife:appraisal a skos:Concept ;
     skos:prefLabel "Appraisal"@en ;
     skos:topConceptOf plife:conceptScheme .
 
+plife:closed a skos:Concept ;
```
geoderekh: these concepts need to be added to the top concept list at the top, yes?

KellyVance:
@geoderekh good pick-up
now also added as topConcept

PR #137, opened 2020-03-20
Update sample-type.ttl

KellyVance:
added notations to concepts without


KellyVance:
The notations should barely ever be seen. 
They are merely there to be used as the prefix code for PIDs

But in this case I have changed the code before it goes into usage.

PR #139, opened 2020-03-20
Update geological-sites.ttl

KellyVance:
Addition of notations

### DIFF
--- a/vocabularies/geological-sites.ttl
+++ b/vocabularies/geological-sites.ttl
@@ -19,6 +19,7 @@
         gsqst:cutting,
         gsqst:field-site,
         gsqst:mine,
+        gsqst:mineral-occurrence
         gsqst:pit,
         gsqst:project-site,
         gsqst:wellbore,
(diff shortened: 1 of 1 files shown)


vocabularies/geological-sites.ttl
```diff
@@ -19,6 +19,7 @@
         gsqst:cutting,
         gsqst:field-site,
         gsqst:mine,
+        gsqst:mineral-occurrence
```
LukeHauck: needs a comma at end.

KellyVance: updated.

vocabularies/geological-sites.ttl
```diff
@@ -114,6 +126,7 @@ gsqst:borehole a skos:Concept ;
         "Well"@en ;
     skos:definition "A Borehole is an intended or actual drilled hole in the ground designed to exchange (or facilitate the exchange of) fluids between a subsurface reservoir and the surface (or another reservoir), or to enable the detection and measurement of rock properties.  Boreholes are the sites of extraction or rock samples, as core or cuttings, and fluid samples. A Borehole has a single Origin and all Wellbore(s) that extend from this Well Origin are parts of the Borehole. Sources: PPDM What Is A Well, GSQ."@en ;
```
GSQ-AI: Does this description bend between status and site? 
As in, a borehole is always a hole bored into the Earth, whereas Intended or Actual are states in which the site finds itself. 
I propose we omit "an intended or actual", to reflect that we're describing the concept of what a borehole is. @KellyVance @DavidCrosswellGSQ @dxwell

DavidCrosswellGSQ: Boy, this definition sure has had some iterations :)

I tend to agree with the omit - is this like Schrodinger's cat?

KellyVance: I suggest omitting "intended or actual", to make the concept definition more robust.
But we should add a scopeNote to explicitly call out that as a regulatory agency for our purposes the borehole exists once it has been proposed, prior to it being physically drilled.

PR #154, opened 2020-03-25
Review (General): Aussie States and Territories

KellyVance:


KellyVance:
@mark-gordon-dnrme have added "Other Territories" concept for completeness. It ill probably be seldom used, but we can always expand it to the individual territories in the future if needed.

KellyVance:
At this point in time I think it is appropriate to limit this to self-governed states and territories. We can always add in other territories if it becomes an issue.

PR #155, opened 2020-03-25
Review (General): Aus Area Codes

KellyVance:


KellyVance:
As per states and territories this is currently limited this to self-governed states and territories. Can expand if this becomes relevant.

KellyVance:
@nicholascar 
changed to http://linked.data.gov.au/def/australian-phone-area-codes/ with prefLabel "Australian Telephone Area Codes"

nicholascar:
> changed to http://linked.data.gov.au/def/australian-phone-area-codes/ with prefLabel "Australian Telephone Area Codes"

How have you changed this? I should see a commit against the file with the corresponding URI/label changes.

PR #158, opened 2020-04-01
Review (General): Qld Geo Features

KellyVance:
Initial GeoFeature _Instances_ vocab
We may also want to incorporate ASUD content in a soon-to-follow edition 2. But not sure the best pathway to do this. Seems like an RDF version of the ASUD should be something GA own.

### DIFF
--- a/vocabularies/queensland-geological-structure-framework-features.ttl
+++ b/vocabularies/queensland-geological-structure-framework-features.ttl
@@ -1,0 +1,730 @@
+@prefix dc: <http://purl.org/dc/elements/1.1/> .
+@prefix dct: <http://purl.org/dc/terms/> .
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix qgsff: <http://linked.data.gov.au/def/queensland-geological-structure-framework-features/> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix sdo: <http://www.schema.org/> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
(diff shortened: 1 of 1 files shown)


johnmckellar:
1.	Broader in each of the two cases below ‘should be new-england-orogen’, not ‘marlborough-province’?

qgsff:new-england-orogen-extent a skos:Concept ;
skos:broader qgsff:marlborough-province ;

qgsff:silverwood-province a skos:Concept ;
skos:broader qgsff:marlborough-province ;

2.	Not sure why the ‘Whitsunday Province’ is assigned to the collections of ‘jurassic-cretaceous-basins’? I see that it is assigned to this group of basins in the Structural Framework, but it does not make good sense. It is a Large Igneous Province, although its beginning, which signifies rifting along the Queensland coast, is represented by the Grahams Creek Formation (preciously the Grahams Creek Grahams Volcanics) at the base of the [latest Jurassic(?) – Early Cretaceous] Maryborough Basin and scattered intrusions further south (like the one on Noosa hill). However, this issue is something to sort out later. We can just let it go for now!!

KellyVance:
@johnmckellar 
well spotted. fixed the broaders (point 1)
point 2 i'll let you sort that out amongst the people involved in the structural framework.. we can change this to reflect any changes you manage to get accepted at the time.

johnmckellar:
OK, thanks Vance

KellyVance:
@DavidCrosswellGSQ An instance list was needed by SRA for selecting valid entities. As @nicholascar says instance data doesn't really make a lot of sense as a skos vocabulary. Are SRA _et al._ now successfully using the [FOI dataset](https://github.com/geological-survey-of-queensland/qldgeofoi-dataset.) published by Nick, to extract the required values and information? 

If so are you comfortable that we close this pull request

DavidCrosswellGSQ:
Yes, the dataset was exported happily for SRA.
I copied the script data_to_csv.py to export the boundary instead of the bounding box geometry.

PR #159, opened 2020-04-01

KellyVance:
Addition of depth-interval to accommodate downhole interval observations as discussed.

### DIFF
--- a/vocabularies/geological-observation-type.ttl
+++ b/vocabularies/geological-observation-type.ttl
@@ -22,6 +22,7 @@
         obtyp:colour,
         obtyp:components,
         obtyp:composition,
+        obtype:depth-interval,
         obtyp:depths-of-supergene-alteration,
         obtyp:fabric,
         obtyp:fossils,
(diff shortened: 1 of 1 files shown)


vocabularies/geological-observation-type.ttl
```diff
@@ -115,6 +113,13 @@ obtyp:composition a skos:Concept ;
     skos:prefLabel "Composition"@en ;
     skos:topConceptOf <http://linked.data.gov.au/def/geological-observation-type>  .
 
+obtyp:depth-interval a skos:Concept ;
+    skos:definition "The depth of an interval or horizon in a borehole."@en ;
```
geoderekh, line 117: A horizon and an interval are fundamentally different, with the latter being bound by the former at the top and base. 
I suggest we remove horizon from the def and later/if necessary, we can add depth-horizon as another concept.

KellyVance: Observation has top and bottom field
for a stratigraphic pick just use the top
for a lithological interval use top and bottom
You can infer that one is a horizon or an interval from the presence of a bottom value.
We can certainly add a horizon observation, i'm just not sure it is necessary with the way the DB is configured.

The observations removed would all be promoted to site level. 
e.g. The site would be a lithological interval. The observation is the depth interval it occurs at. 
It would be duplication to have each interval as both a site type and an observation type

You either have each interval type as a site so you can link directly to the feature then define its depth in the observation.
or
You have a generic wellbore-interval site type and make each different interval a different observation type. 

But the former allows you to have the depth observation (or observations) on site type X along with other observations, and allows to to hang results off each interval too. 
e.g. have a single reservoir interval site, 
the reservoir is made of two sand packages, each with individual depth observations. 
Each observed depth interval is then linked to independent results such as porosity and permeability etc.

KellyVance:
Based on recent discussions and re-configuration of model this branch is redundant and new pull requests and reviews will be needed.

PR #163, opened 2020-04-06
Update result-type.ttl

KellyVance:
Adding in angle types with specific absolute references and context
Azimuth
Inclination (what Petroleum use)
Declination (what Minerals use)

Just a thought. Feel free to correct, reject, edit, etc.

### DIFF
--- a/vocabularies/result-type.ttl
+++ b/vocabularies/result-type.ttl
@@ -18,6 +18,7 @@
         rslt:age,
         rslt:api-gravity,
         rslt:area,
+        rslt:azimuth,
         rslt:capacitance,
         rslt:casing-weight,
         rslt:colour,
@@ -28,6 +29,7 @@
         rslt:cost,
         rslt:count,
         rslt:data,
+        rslt:declination,
         rslt:density,
         rslt:descriptive,
         rslt:distance,
@@ -49,6 +51,7 @@
         rslt:heat,
         rslt:illuminance,
         rslt:inductance,
+        rslt:inclination,
         rslt:length,
         rslt:luminous-flux,
         rslt:luminous-intensity,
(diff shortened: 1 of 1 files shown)


geoderekh:
Not quite right with the comment there, though I'm out of energy to edit the concepts this afternoon. 

Declination is one altLabel for azimuth (0-360) although it is also used to express the angle between of magnetic and true north (+ive/-ive or E/W). There are also a bunch of conventions for azimuth notation (strike, dip/plunge direction, quadrant) but I'd like to force everyone to adopt dip-dipdirection because it is really best practice. 

Inclination is horizontal to vertical. Plunges (linear features) or dips (planar features) have inclination. This is where petroleum like to be different - they measure inclination from the vertical, whereas all structural geology measurements and also minerals boreholes measure it from the horizontal. 
Then there is the problem of upward inclinations, which are -90>0 for most structural geology, 90>180 for petroleum boreholes, and 0>90 for disciplines other than earth science. 

Depending on the domain scope for qudt this might be troublesome to write.

KellyVance:
There is an issue that these terms are used in incredibly variable ways. Virtually every possible variation of interpretation is true to someone. But we need to land on how we refer to things.

Definition of **declination**
1: angular distance north or south from the celestial equator measured along a great circle passing through the celestial poles
4: a bending downward : INCLINATION
6: the angle formed between a magnetic needle and the geographical meridian

Definition of **inclination**
2a: a deviation from the true vertical _or_ horizontal : SLANT

I would suggest azimuth may need subconcepts for strike and dip-direction. 
azimuth for things not strictly geological
strike for when we don't know the dip direction
dip direct for whenever we do know and/or can enforce into the future

For vertical angles we can either convert everything stored in the Db as it enters to common framework, or pick one term to define departure from vertical, and one to define departure from horizontal. Really unsure the best way here other than that if we want to store these things we need a consistent way to do so.

PR #170, opened 2020-04-09
Update result type

### LINKED ISSUE #168: Magnetic Susceptibility missing from Units of Measure.
Hi Vance.  The units of measure seems to be missing Magnetic Susceptibility SI (unitless)

KellyVance:
Introduce concept of Magnetic Susceptibility as a result type.
measurement is unitless so UoM will be null/blank/not applicable/unitless.

Also rename Dog Leg Severity (industry specific) to Curvature (general)

closes #168

### DIFF
--- a/vocabularies/result-type.ttl
+++ b/vocabularies/result-type.ttl
@@ -27,11 +27,11 @@
         rslt:concentration,
         rslt:cost,
         rslt:count,
+        rslt:curvature,
         rslt:data,
         rslt:density,
         rslt:descriptive,
         rslt:distance,
-        rslt:dog-leg-severity,
         rslt:drill-rate,
         rslt:electric-charge,
         rslt:electric-current,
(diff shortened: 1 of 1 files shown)


DavidCrosswellGSQ:
> measurement is unitless so UoM will be null/blank/not applicable/unitless.
What are the permissible result values for this result type?

Greenwoodmatthew:
@DavidCrosswellGSQ Mag sus is a ration of the magnetisation to the applied field.  common rock values are often in the 10^-5 - 10^-3 range but can be very high (iron and other para-magnetic minerals) or even negative

KellyVance:
@Greenwoodmatthew added alternate term for capacitance/
rslt:capacitance a skos:Concept ;
    skos:altLabel "Chargeability"@en ;

DavidCrosswellGSQ:
@Greenwoodmatthew so would the unit of measure then be ratio? or magnetism per unit volume?
Just trying to understand why there would not be a unit of measure.

DavidCrosswellGSQ:
anyhoo - the question is unrelated to the approval of this vocab and can be answered very much later

Greenwoodmatthew:
Thanks Vance.  I am happy with changes

KellyVance:
Thanks @Greenwoodmatthew 
As a matter of procedure can you please: Files Changed>Review Changes>Approve>Submit Review?

PR #171, opened 2020-04-09
Delete seismic-sampling-method.ttl

KellyVance:
delete seismic sampling method
vocabulary integrated into survey method 'svymh:seismic-methods a skos:Collection'


KellyVance:
No worries. I'll close this pull request but raise a separate issue so it is documented for a later action.

PR #181, opened 2020-04-14
Removed nesting of Boundary concepts

DavidCrosswellGSQ:
Flattened scheme to make it suited for drop-downs. Cleaned up code.

### DIFF
--- a/vocabularies/geometry-roles.ttl
+++ b/vocabularies/geometry-roles.ttl
@@ -13,14 +13,17 @@
     a owl:Ontology , skos:ConceptScheme ;
     dct:creator <http://linked.data.gov.au/org/gsq> ;
     dct:created "2020-03-12"^^xsd:date ;
-    dct:modified "2020-03-12"^^xsd:date ;
+    dct:modified "2020-04-14"^^xsd:date ;
     dct:publisher <http://linked.data.gov.au/org/gsq> ;
     skos:definition """A Geometry Role is the function a Geometry plays in relation to the Feature it is a Geometry for where Feature and Geometry are defined by the GeoSPARQL Ontology.
     
 Geometry Role is not geometry type and does not indicate sub classes of geometry. Well known Geometry types or subclasses of geometry include Point, Polygon, LineString MultiPolygon etc."""@en ;
     skos:hasTopConcept 
         grole:centroid ,
-        grole:boundary ,
+        grole:bounding-box ,
+        grole:bounding-circle ,
+        grole:concave-hull ,
+        grole:convex-hull ,
         grole:detailed-geometry ;
     skos:prefLabel "Geometry Role"@en ;
     skos:scopeNote "Concepts in this vocabulary indicate should, but do not have to only be, used with instances of the GeoSPARQL Ontology's Geometry class using the GeoSPARQL Extensions Ontology's hasRole property."@en .
@@ -28,8 +31,7 @@ Geometry Role is not geometry type and does not indicate sub classes of geometry
 <http://linked.data.gov.au/org/gsq>
     a sdo:Organization ;
     sdo:name "Geological Survey of Queensland" ;
-    sdo:url <https://www.business.qld.gov.au/industries/mining-energy-water/resources/geoscience-information/gsq> ;
-.
+    sdo:url <https://www.business.qld.gov.au/industries/mining-energy-water/resources/geoscience-information/gsq> .
 
 grole:centroid a skos:Concept ;
     dct:source <https://en.wikipedia.org/wiki/Centroid> ;
@@ -37,11 +39,9 @@ grole:centroid a skos:Concept ;
     skos:definition "The arithmetic mean position of all the Points in a Feature's Geometry, expressed as a single Point."@en ;
     skos:hiddenLabel "center of mass"@en ;
     skos:inScheme <http://linked.data.gov.au/def/geometry-roles> ;
-    skos:prefLabel "Centroid"@en ;
-    skos:topConceptOf <http://linked.data.gov.au/def/geometry-roles> .
+    skos:prefLabel "Centroid"@en .
 
 grole:bounding-box a skos:Concept ;
-    skos:broader grole:boundary ;
     dct:source <https://en.wikipedia.org/wiki/Minimum_bounding_box> ;
     skos:altLabel "minimum bounding box"@en , "smallest bounding box"@en , "enclosing box"@en , "bounding rectangle"@en ;
     skos:definition "The box (rectangle) with the smallest area that contains all the Points in a Feature's Geometry."@en ;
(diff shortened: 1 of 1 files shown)


KellyVance:
@DavidCrosswellGSQ @nicholascar  Pull request one month old. Still an active issue? Or should this request be closed and its associated branch be delted?

DavidCrosswellGSQ:
I don't mind either way - I was trying to simplify this vocab.
I think that the current vocab may confuse - when do I use a boundary vs detailed geometry?

vocabularies/geometry-roles.ttl
```diff
@@ -13,70 +13,59 @@
...
 
 grole:convex-hull a skos:Concept ;    
     dct:source <https://www.arcgis.com/home/item.html?id=a6129839e96d4c0682250673cd89589d> ;
-    skos:broader grole:boundary ;
     skos:definition "The minimum bounding Polygon encompassing all the all Points in a Feature's Geometry."@en ;
     skos:inScheme <http://linked.data.gov.au/def/geometry-roles> ;
     skos:prefLabel "Convex hull"@en .
 
 grole:detailed-geometry a skos:Concept ;
     skos:definition "A Feature's Geometry in detail. It may be regular or irrigular but it does not constitute a Boundary for the Feature."@en ;
```
KellyVance, line 69: "irrigular" typo

KellyVance:
@nicholascar can you please comment on the below?

I think perhaps to increase specificity, broaden the scope, and decrease the confusing terms we need a bit more structure with something akin to:

- Boundary (aka Polygon)
    - Bounding Box
    - Bounding Circle
    - Bounding Ellipse
    - Convex Hull
    - Concave Hull
    - Detailed Polygon
- Line
    - Simple Line
    - Polyline
    - Multiline
- Point
    - Centroid
    - Precise Location
- Composite
    - Complex Geometry (combination of points +/- lines +/- polygons) 

- Collection: detailedGeometry
    members: detailed polygon, polyline, multiline, precise location, complex geometry

- Collection: forDisplay
    (all the non-TopConcepts)

nicholascar:
> can you please comment on the below?

Looks sensible to me, as in whatever you want, you shall get and I have no objections to that hierarchy except:

* you don't need a *forDisplay* collection: just ensure that when you do display these things, you select terms that are not topConceptOf. This is akin to the [filtering of ESDC for the 2 top levels, as per the Vocabs wiki](https://github.com/geological-survey-of-queensland/vocabularies/wiki/Vocabulary-SPARQL-queries#8-get-the-top-2-levels-of-a-vocab) with a slightly altered query logic. 
* longer term, we will have to unpick the Polygon/Point/Line classification v. Roles.
   * I think GSQ should put a submission into the OGC's GeoSPARQL revision WG (which I'm involved with) to ask them to solve this for everyone.
      * does the OGC recommend more types of geometry (e.g. Poylgon & sub-classes such as Convex Hull) or does it recommend a type/role split?

geoderekh:
I'm coming in late on this one, from a geoproperties perspective, we need both type and role, however role isn't in the PostGIS schema. It is in SPATIAL.ROLE, wheres type (point, line, multipoint etc) is part of the PostGIS extension in GEOMETRY_COLUMNS.TYPE. 

In PostGIS v3.0 type is restricted to POINT, LINESTRING, POLYGON, MULTIPOINT, MULTILINESTRING, MULTIPOLYGON, GEOMETRYCOLLECTION or corresponding XYM versions POINTM, LINESTRINGM, POLYGONM, MULTIPOINTM, MULTILINESTRINGM, MULTIPOLYGONM, GEOMETRYCOLLECTIONM. This attribute is required for ensuring type homogeneity. 

Vance's solution could work where our subclasses (true roles) could be resolved to role whereas the broader class could resolve for the type classification needed for postgis. Depending on how SRA built the CRUD app (i.e. if GEOMETRY_COLUMNS.TYPE is user-specified or inferred), we may even *need* to use it so we can set both type and role in the one field. 

Going forward, I think there is an obvious need for a distinct attribute of roles and would be happy to present the case to the OGC if someone points me in the right direction.

KellyVance:
This may need an overhaul. We may need a new Geometry Type vocabulary (even if we don't need to use it) just so there is a clear delineation between GIS compatible _types_ and the role of how a shape represents its subject. i.e. is it an approximation of the actual shape or location (detailed), a weighted point representing a larger feature (centroid), or some kind of simplified generalised shape of the subject (the rest). Thoughts?

Type | Role
-- | --
POINT | Detailed Geometry
LINESTRING | Centroid
POLYGON | Boundary(dont display)
MULTIPOINT | Bounding Box
MULTILINESTRING | Bounding Circle
MULTIPOLYGON | Bounding Ellipse
GEOMETRYCOLLECTION | Hull

vocabularies/geometry-roles.ttl
KellyVance, line 69:
```suggestion
    skos:definition "A Feature's Geometry in detail. It may be regular or irregular. A detailed geometry is an exact representation, or close approximation of the features true shape. Distinct from a boundary, which is a generalised representation."@en ;
```

vocabularies/geometry-roles.ttl
```diff
@@ -13,70 +13,59 @@
...
     skos:definition "The box (rectangle) with the smallest area that contains all the Points in a Feature's Geometry."@en ;
     skos:inScheme <http://linked.data.gov.au/def/geometry-roles> ;
     skos:prefLabel "Bounding box"@en .
 
 grole:bounding-circle a skos:Concept ;
-    skos:broader grole:boundary ;
     dct:source <https://en.wikipedia.org/wiki/Smallest-circle_problem> ;
     skos:definition "The circle with the smallest area that contains all the Points in a Feature's Geometry."@en ;
     skos:inScheme <http://linked.data.gov.au/def/geometry-roles> ;
     skos:prefLabel "Bounding circle"@en .
```
KellyVance, line 56:
```suggestion

grole:boundary a skos:Concept ;
    skos:definition "A polygonal shape that approximates the extent of a Feature's Geometry through a simplified and generalised shape e.g. representing a complex array of lines as a bounding box. Distinct from a detailed geometry as it lacks the detail and specificity to describe the true geometry of the feature."@en ;
    skos:inScheme <http://linked.data.gov.au/def/geometry-roles> ;
    skos:prefLabel "Boundary"@en ;
    skos:topConceptOf <http://linked.data.gov.au/def/geometry-roles> .


```

vocabularies/geometry-roles.ttl
geoderekh: added the words 'enclosing the feature' for clarity. a centroid is a generalised representation  but is not a boundary
```suggestion
    skos:definition "A Feature's Geometry in detail. It may be regular or irregular. A detailed geometry is an exact representation, or close approximation of the features true shape. A detailed geometry may correspond to any geometry type (e.g. point, line, polygon) appropriate to the specific feature being described at the scale of representation. Distinct from a boundary, which is a generalised representation enclosing the feature."@en ;
```

vocabularies/geometry-roles.ttl
geoderekh: slight tweak to clarify 
```suggestion
    skos:definition "A simplified and generalised polygonal shape that encloses the detailed geometry of a feature and approximates the feature extent e.g. representing a complex array of lines as a bounding box. Distinct from a detailed geometry as it lacks the detail and specificity to describe the true geometry of the feature."@en ;
```

PR #182, opened 2020-04-14
Made Driller a top concept

DavidCrosswellGSQ:
Removed Driller as a narrower concept of Contractor

### DIFF
--- a/vocabularies/gsq-roles.ttl
+++ b/vocabularies/gsq-roles.ttl
@@ -10,12 +10,13 @@
 <http://linked.data.gov.au/def/gsq-roles> a owl:Ontology , skos:ConceptScheme ;
     dct:created "2019-10-17"^^xsd:date ;
     dct:creator <http://linked.data.gov.au/org/gsq> ;
-    dct:modified "2020-01-10"^^xsd:date ;
+    dct:modified "2020-04-14"^^xsd:date ;
     dct:publisher <http://linked.data.gov.au/org/gsq> ;
     dct:source <https://standards.iso.org/iso/19115/resources/Codelists/gml/CI_RoleCode.xml> ;
     skos:definition "The roles of Agents (People and Organisations) in relation to Entities (Datasets, Samples etc.)"@en ;
-    skos:hasTopConcept gsqr:contractor,
+    skos:hasTopConcept 
         gsqr:contractor,
+        gsqr:driller,
         gsqr:operator,
         iso11179:Submitter,
         isorole:author,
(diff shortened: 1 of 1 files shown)


nicholascar:
As per all other comments: don't reduce the information content of the vocabs: find better ways to view (query) the data as needed. 

Changes to simplify things here will only set a precedent that you should do bad modelling to suit some simple purpose.

DavidCrosswellGSQ:
@nicholascar while I understand your comments, I don't think that is the issue here.
driller is not a narrower term of contractor.
a driller might not be a contractor.
Also, there is an error in this vocab that is fixed in this PR - duplication of contractor in Top Concepts.

geoderekh:
Maybe I'm wrong, but it seems having contractor, operator, and driller all as top concepts, would imply that drillers are never contractors or operators. 

We will have instances where driller would be either contractor->driller (a drilling subcontractor) or operator->driller (staff drillers/own rigs). The distinction could be a valuable metric for evaluating factors affecting exploration success, so it might be nice to have two 'driller' concepts instead of one.

nicholascar:
If it gets hard to work out what something is Driller and/or Contractor then I suppose you’d best have separate Concepts. You could dual tag an Agent as both a Contractor and a Driller.... I assume you *don’t* want to allow that in a UI but data wise multiple roles would be fine.

DavidCrosswellGSQ:
In the geoprops model, we have one entry for Operator - in Borehole header table, and one entry for driller in the Wellbore table. This could be the same party or different parties.

If you want to know if someone is a contractor - then we could record that elsewhere - a simple contractor register.

image

KellyVance:
@nicholascar @DavidCrosswellGSQ 

A driller may be an operator or a contractor.
But it begs the question of whether we need driller as a role concept at all if the data model itself defines the drilling 'role'. The entry in the _driller_ field is-and-can-only-be a driller by definition.

Then the role in the org table becomes whether the operator is doing the driller or a contractor is doing the drilling.

KellyVance:
@nicholascar @DavidCrosswellGSQ  PR one month old. Do we have a resolution?

DavidCrosswellGSQ:
@KellyVance I still stand by this change and would like it approved.

nicholascar:
Not for me to decide, sorry! 

You can always model soemthing with a direct property:

```
<a> eg:hadDriller <b> .
```

Or with a general relation + role:

```
<a> eg:hadRelatedAgent <b> .

<b> eg:hadRole <Driller> .
```

And the choice can be semantically neutral (i.e. you can entail one from the other) so it's really just a style thing. If the specific property, `eg:hadDriller`, would be well understood and common you could use that. If the Agent + Role pattern would be generally applicable, like for lots of potential roles + a common vocab, you could use that.

KellyVance:
I'm still a little unsure about the implications. But if you want to reopen this request I'm comfortable approving this and adjusting if/when it breaks.

PR #183, opened 2020-04-14
Flattened the vocab

DavidCrosswellGSQ:
Hi @geoderekh 
I have moved the narrower terms out from under "follow-up".
So, now all terms are top concepts.

### DIFF
--- a/vocabularies/survey-relationship-type.ttl
+++ b/vocabularies/survey-relationship-type.ttl
@@ -13,10 +13,13 @@
         skos:ConceptScheme ;
     dct:created "2020-04-08"^^xsd:date ;
     dct:creator <http://linked.data.gov.au/org/gsq> ;
-    dct:modified "2020-02-20"^^xsd:date ;
+    dct:modified "2020-04-15"^^xsd:date ;
     dct:source "Developed by the Geological Survey of Queensland." ;
     skos:definition "The relationships that exist between surveys."@en ;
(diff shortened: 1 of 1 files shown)


DavidCrosswellGSQ:
@geoderekh so do we change the preflabels to be prefixed with "Follow-up"?

nicholascar:
No no no! Absolutely not 

> do we change the preflabels to be prefixed with "Follow-up"?

You're introducing an anti-pattern here by removing structure from data and placing it into a "micro-format" of label names. 

As per my other comments on flattening, you shouldn't be doing this at all.

I gather that all the moves to flatten are motivated by a need for SRA to get lists of Concepts simply. We'll we have query-able graphs here so we can work out a better way than just dumbing down the vocabs. OrderedCollections or queries etc. If really necessary, an API filtering end-point.

geoderekh:
We need to resolve the larger flattening issue today as we need to approve the GeoProps Data Elements by COB. 
I think the simplest solution is to leave the vocabs as they are and for SRA to just get all concepts where we might have specified in the Data Elements v1.6 to get all top concepts or all narrower concepts. 
SRA will still need to improve their code to implement the conditional logic based on collections or attributes (where necessary), but the balance of the lookups could be handled in a consistent fashion.

DavidCrosswellGSQ:
No change to the vocab required.

PR #184, opened 2020-04-20
removed narrower survey site types, renamed project site to survey site

geoderekh:
As discussed on 20/4
Seismic and geophysical survey areas aren't really distinct typs of sites, they are sites where a survey was conducted, with geophysics/seismic coming from the survey type. 
Project site renamed to survey site.

### DIFF
--- a/vocabularies/geological-sites.ttl
+++ b/vocabularies/geological-sites.ttl
@@ -22,7 +22,7 @@
         gsqst:mineral-occurrence,
         gsqst:mineral-deposit,
         gsqst:pit,
-        gsqst:project-site,
+        gsqst:survey-site,
         #gsqst:wellbore,
         gsqst:wellbore-interval ;
     skos:prefLabel "Geological Sites"@en .
@@ -45,13 +45,6 @@ gsqst:colluvial a skos:Concept ;
     skos:notation "cv" ;
     skos:prefLabel "Colluvial Site"@en .
 
-gsqst:geophysical-survey-area a skos:Concept ;
-    skos:broader gsqst:project-site ;
-    skos:definition "An area over which a non-seismic geophysical survey is conducted."@en ;
-    skos:inScheme <http://linked.data.gov.au/def/geological-sites> ;
-    skos:notation "ga" ;
-    skos:prefLabel "Geophysical Survey Area"@en .
-
 gsqst:outcrop a skos:Concept ;
     skos:broader gsqst:field-site ;
     dct:source "Merriam-Webster Dictionary" ;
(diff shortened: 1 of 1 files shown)


geoderekh:
Good pickup Andrew! 

The minerals usage of the term is a valid business requirement, but I don't recall why it was added in here as a type of site. 

A project is more like a relationship between or a collection of entities across the different sosa levels. It is the rationale that ties together a bunch of activities (surveys and observations) along with all of the derivative things and data that go together to inform the project outcomes. 

While sites can contain other sites, once the site is 'big' enough to no longer be just a proxy for the feature of interest, but rather to be the entire feature of interest itself (and the project area is, by definition, the area of interest of a project), the entity would be a Feature. 

A project is therefore the ultimate FOI for those involved and because investigators are ultimately interested in the geological properties of the Project Area, it needs to be able to link to all of the sites, surveys, samples, observations, and results of relevance. Feature fits our needs perfectly I think. 

@CourteneyD please correct me if I'm wrong here.

CourteneyD:
> Good pickup Andrew!
> 
> The minerals usage of the term is a valid business requirement, but I don't recall why it was added in here as a type of site.
> 
> A project is more like a relationship between or a collection of entities across the different sosa levels. It is the rationale that ties together a bunch of activities (surveys and observations) along with all of the derivative things and data that go together to inform the project outcomes.
> 
> While sites can contain other sites, once the site is 'big' enough to no longer be just a proxy for the feature of interest, but rather to be the entire feature of interest itself (and the project area is, by definition, the area of interest of a project), the entity would be a Feature.
> 
> A project is therefore the ultimate FOI for those involved and because investigators are ultimately interested in the geological properties of the Project Area, it needs to be able to link to all of the sites, surveys, samples, observations, and results of relevance. Feature fits our needs perfectly I think.
> 
> @CourteneyD please correct me if I'm wrong here.

No this makes sense but I don't understand how we are going to be recording information at the Feature level to know if that is where this 'Project' level info sits.

vocabularies/geological-sites.ttl
```diff
@@ -194,12 +172,12 @@ gsqst:wellbore-interval a skos:Concept ;
...
 
-gsqst:project-site a skos:Concept ;
+gsqst:survey-site a skos:Concept ;
     skos:altLabel "Project"@en ;
     skos:definition "The location of a survey."@en ;
     skos:inScheme <http://linked.data.gov.au/def/geological-sites> ;
-    skos:notation "pj" ;
-    skos:prefLabel "Project Site"@en ;
+    skos:notation "ss" ;
+    skos:prefLabel "Survey Site"@en ;
```
KellyVance, line 180: "ss" is the prefix for survey type: seismic survey
Prefixes (notations) should be unique throughout the ecosystem of features, site, survey, sample, obs, results etc.
Helps to ensuring PID distinctiveness. Since it is for a machine read purpose (rather than human interrogation) there is no harm leaving the notation as pj for uniqueness and leaving project site as an alternative label. Otherwise this makes sense to amalgamate project/survey sites to incorporate seismic surveys and let the survey type deal with the survey type.

geoderekh: Good pick up @KellyVance, though we should be k because there are no more seismic surveys, just geophysical surveys. We need to get you up to speed on the survey front since we finally got to the bottom of surveys in sosa and have a new way to store "multi-type" surveys.

KellyVance: Thats fine. we will have to edit the survey type. but in the meantime still a good idea to keep notations unique at any point in time. We should push this along and get the approvals to cclose out this pull request

KellyVance:
duplicate of #194

PR #185, opened 2020-04-20
deprecated 4 narrower concepts to normalise vocab

geoderekh:
This vocab contains many top concepts but hardly any narrower concepts.  Ideal state is for more granularity across all of the survey types but this is a large body of work that has not been completed. Until then it should stay relatively flat. Concepts just #'d out for easy reinstatement in the future.


LukeHauck:
No problem with a mostly flat list of top concepts and a few narrowers.
Speaking for Seismic, those hashed out comments related to Passive Seismic are covered by the seismic sampling method vocab. In that, we treat it as a seismic survey with a passive energy source, rather than a distinct type of survey as detailed here. As far as I'm aware, other geophysics and geochemistry do not have a vocab for what's hashed out, so should stay here for now until this is better fleshed out.

DavidCrosswellGSQ:
"vestigial limb" becomes part of the vocab vernacular

PR #201, opened 2020-05-06
removed -datum

nicholascar:
I've removed the "-datum" part of the URI and Datum from the vocab name since not all of the depth references contained are Datums: some are reference points.

I think `/def/depth-reference` will work much better as a vocab URI as it would be open to other references and doesn't indicate a restriction to datums that isn't supported by the vocab.

### DIFF
--- a/vocabularies/depth-reference.ttl
+++ b/vocabularies/depth-reference.ttl
@@ -1,14 +1,10 @@
-@prefix dc: <http://purl.org/dc/elements/1.1/> .
 @prefix dct: <http://purl.org/dc/terms/> .
-@prefix drd: <http://linked.data.gov.au/def/depth-reference-datum/> .
+@prefix drd: <http://linked.data.gov.au/def/depth-reference/> .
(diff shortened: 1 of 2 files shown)


vocabularies/depth-reference.ttl
```diff
@@ -24,17 +20,17 @@
         drd:seismic-reference-datum,
         drd:water-bottom,
         drd:well-head ;
-    skos:prefLabel "Depth Reference Datum"@en .
+    skos:prefLabel "Depth Reference"@en .
     
 drd:absolute a skos:Collection ;
-    skos:definition "A datum that describes an absolute plane of reference at a global or regional scale."@en ;
+    skos:definition "A reference that describes an absolute plane of depth at a global or regional scale."@en ;
```
KellyVance, line 26: To be consistent with change definition should be to describe an absolute plane _or point_
More accurately "A fixed plane or point that describes an absolute reference for depth observations.

PR #202, opened 2020-05-07
Telephone Type alignment with VCARD

nicholascar:
Alignment of the telephone types with VCARD, a W3C standard. This is simple URI reuse, no definitional change. The odd diff is because the file was renamed in a separate commit to the edit. Really, no changes except for renaming and URI replacement.

Even though this vocab is derived from AS4590, there is not likely to be an authoritative vocab publication of AS4590 codes just yet (hopefully, next Financial Year as I work on the National Data Commission) so we're best off aligning internationally and picking up AS4590 when they appear.


vocabularies/telephone-type.ttl
```diff
@@ -0,0 +1,41 @@
...
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+
+<http://linked.data.gov.au/def/telephone-type> a owl:Ontology,
+        skos:ConceptScheme ;
+    dcterms:created "2020-03-25"^^xsd:date ;
+    dcterms:creator <http://linked.data.gov.au/org/gsq> ;
+    dcterms:modified "2020-03-25"^^xsd:date ;
+    dcterms:publisher <http://linked.data.gov.au/org/gsq> ;
+    dcterms:source "Derived from Australian Standard AS 4590.1:2017" ;
```
KellyVance, line 16: If we are now referencing  VCARD our source should probably indicate this.

vocabularies/telephone-type.ttl
```diff
@@ -0,0 +1,41 @@
+@prefix dcterms: <http://purl.org/dc/terms/> .
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix sdo: <http://www.schema.org/> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
+@prefix tsl: <http://linked.data.gov.au/def/telephone-type/> .
+@prefix vcard: <http://www.w3.org/2006/vcard/ns#> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+
+<http://linked.data.gov.au/def/telephone-type> a owl:Ontology,
```
KellyVance, line 10: How important is aligning the URI to the prefLabel?
If we keep it the same URI there is no need to recode the vocabulary references currently linked to this.
It's the lazy option, I'm happy to go either way, but i'd rather not create work elsewhere if i don't need to.

vocabularies/telephone-type.ttl
```diff
@@ -0,0 +1,41 @@
...
+
+<http://linked.data.gov.au/org/gsq> a sdo:Organization ;
+    sdo:name "Geological Survey of Queensland" .   
+
+vcard:Cell a skos:Concept , vcard:TelephoneType ;
+    skos:altLabel "Cell"@en,
+        "Cellular"@en ;
+    skos:definition "A portable telephone that can make and receive calls over a radio frequency link while the user is moving within a telephone service area."@en ;
+    skos:inScheme <http://linked.data.gov.au/def/telephone-type> ;
+    skos:notation "01" ;
```
KellyVance, line 31: These notations were from the AS4590 standard.
If our source now is vcard are the notations still valid?

vocabularies/telephone-type.ttl
```diff
@@ -0,0 +1,41 @@
...
+    skos:prefLabel "Mobile"@en ;
+    skos:topConceptOf <http://linked.data.gov.au/def/telephone-type> .
+
+vcard:Voice a skos:Concept ;
+    skos:altLabel "Landline"@en , "voice"@en ;
+    skos:definition "A telephone that can make and receive calls by transmission of communications signals via a cable to a fixed location."@en ;
+    skos:inScheme <http://linked.data.gov.au/def/telephone-type> ;
+    skos:notation "02" ;
```
KellyVance, line 39: as above re: notations

vocabularies/telephone-type.ttl
nicholascar: No, I don't think so. `dcterms:source` should be used to indicate any content, RDF or otherwise, from which this vocab is delivered however in this case we aren't *deriving* from VCARD but rather doing the normal RDF thing of reusing VCARD elements, just as we reuse SKOS and OWL elements, so I think no source property is necessary.

We could use source if there was no other way to know where content came from but with RDF reuse, we have the URIs.

vocabularies/telephone-type.ttl
nicholascar: > How important is aligning the URI to the prefLabel?

It's not required and sometimes can't be achieved. There can be no formal SKOS or other rule to mandate such a thing (you would need regular expression checking) but even if there was, it would defeat a SKOS purpose which is to separate identify (URI) from labels so that different situations can be catered for. 

I think its *required* that labels reflect the intention of the concept and *preferable* that they echo the URI. Sometimes you have numerical URIs (e.g. http://purlorg/obo/0_12345) that clearly can't have URIs echoing them.

vocabularies/telephone-type.ttl
nicholascar: Sure. Notation is there to allow you to associate a term with something else, such as a code used in a particular environment. You might have to add a note as to what that environment is, perhaps in the ConceptScheme scopeNote you could say this.

nicholascar:
I'm arranging for the Federal govt committee IT-004, run by Standards Australia, that "owns" AS4590 to put out authoritative forms of all AS4590 vocabs. So we should use this one until they become available and then cut across to using them.  

I'll likely be the person actually making the AS4590 SKOS versions...

vocabularies/telephone-type.ttl
KellyVance: I'm happy to approve as-is and let the devs know of the redirection.

PR #209, opened 2020-05-11
Review - Geological Observation Instruments

KellyVance:
Edition One of Geological Observation Instruments
Includes Seismic i.e. Geophones and Aerial Geophysics

Content review @Greenwoodmatthew @CantRoger @LukeHauck 
Vocabulary is known to be incomplete.
Purpose is to have an active vocabulary that can be updated and added to as people contribute additional information.

@nicholascar please run SHACL validation.


KellyVance:
> Line 400, HeliSAM is an EM method.
> I like this vocab, sorry for not doing more for you on the definition front Vance.
> Migrating the AG stuff should be possible using this as I presume we can map to either a top concept or one of the more specific instruments as required. Is there an Observation type Vocab anywhere containing the Mag, Rad, EM, Grav… etc?

Moved HeliSam from Mag collection to EM collection.
Correct about mapping to the top concept. If, for example, you know it is a scintrex instrument but not the model it can be mapped to the top concept.
Planning on tidying up survey type, survey method, obs type and obs method today. Will require some assistance, particularly with obs method.

KellyVance:
@DavidCrosswellGSQ  and/or @LukeHauck can you please run a technical check over this pull request? I don't think Nick will have capacity to get to this on his more limited hours now.

smrgeoinfo:
feature suggestion:  seems like having a link to the [observation type](https://github.com/geological-survey-of-queensland/vocabularies/blob/master/vocabularies/geological-observation-type.ttl) would be really useful for constructing user interfaces for people to find an instrument.   Perhaps use skos:related?  A quick looks at geological-observation-type suggests that not all observed properties you'd need are in that vocabulary.

PR #238, opened 2020-06-02
Update earth-science-data-category

KellyVance:
- Reordered concepts to  be in hierarchical order.
- Changed Collection to Ordered Collection for more structured display in portal and reordered entries to reflect main hierarchy (with exception of Geomechanics, which I moved from below engineering into geology because it better suits our specific use case)
- Added Earth Science as a TopConcept for consistency
- Added Geology into collection to make hierarchy more apparent and intuitive.


KellyVance:
> I think spectral-imaging should sit outside geophysics. As a methof spectral core imaging would not be referred to as a geophysical process. I think spectral-imaging should be it's own top concept? Otherwise i think it looks good. Hard to see without it being deployed to UAT and testing in that environment

@Greenwoodmatthew Good point. As per [ANZSRC FOR](https://vocabs.ands.org.au/viewById/20) I have introduced Photogrammetry and Remote Sensing under Geomatic Engineering and slotted the spectral imaging concepts under there. I have also reordered the ordered collection accordingly.

KellyVance:
@Greenwoodmatthew FYI while i'm waiting for final approval on the Pull request I've just done a commit to adjust the concept order in line with the field-of-research order, and added some resources engineering concepts. Feel free to comment if needed, or i'll take silence as your approval still standing.

vocabularies/earth-science-data-category.ttl
```diff
@@ -359,15 +366,8 @@ esdc:geothermics-and-radiometrics a skos:Concept ;
...
-    skos:prefLabel "Geothermics and Radiometrics"@en ;
-    skos:altLabel "Radiometrics"@en .
-
-esdc:geothermal-energy a skos:Concept ;
-    skos:broader esdc:geothermics-and-radiometrics ;
-    skos:definition "The study of heatflow in the Earth's upper crust in the context of energy generation, and methods, materials and operations used in extraction of energy from geothermal sources."@en ;
-    skos:inScheme <http://linked.data.gov.au/def/earth-science-data-category> ;
-    skos:altLabel "Geothermal"@en ;
-    skos:prefLabel "Geothermal Energy"@en .
+    skos:prefLabel "Radiometrics"@en ;
```
Greenwoodmatthew, line 369: I preferred the other terms list.  But in terms of radiometric they way we use it maybe it should be separate as it is not about Geothermics and Radiometrics and heat flow, but surface radioisotopes

KellyVance: [ANZSRC-FOR](https://vocabs.ands.org.au/viewById/20) classifies it as Geothermics and Radiometrics.
If they should be separate I suggest that we introduce them as narrower concepts so we still retain a direct mapping to the established standard.

As a preference I would really like to get this PR approved and start a new one while these edits are still timely and this change stacks up too many commits to easily follow the changes.

Greenwoodmatthew: Yeah i am happy as 80:20 to go,  But on a techincal point i think we are still wrong path for this element

Greenwoodmatthew: For ANZSCR-FOR this is why i think it is not suited to the ESDC as the data categories do not relate to the fields of research often

LukeHauck:
Unexpected """ on line 561.

PR #247, opened 2020-06-04
Temporary UoM Vocabulary for UAT

KellyVance:
A hastily assembled skos Units of measure vocabulary
This is so UAT can continue and select values to input results.

This file should be overwritten or removed once we have the correct vocabulary but this is essential to move forward with UAT and system development.


KellyVance:
No Longer Applicable

PR #248, opened 2020-06-04
Update geological-sites.ttl

KellyVance:
Update to Geological Sites based on feedback from mapping for data migration.
New term _Regolith_
New term _Saprolith_
Altered term _Soil_ -> _Pedolith_
New term _Subcrop_
Term split _Tailings_ -> _Tailings_ and _Spoil_

SME @BlakePaul 
Vocabulary Owner @geoderekh 
Tech Check @DavidCrosswellGSQ or @LukeHauck 

FYI for UAT @lisa-woody


vocabularies/geological-sites.ttl
```diff
@@ -138,32 +143,69 @@ gsqst:colluvial a skos:Concept ;
 
 gsqst:outcrop a skos:Concept ;
     skos:broader gsqst:field-site ;
-    dcterms:source "Merriam-Webster Dictionary" ;
+    dct:source "Merriam-Webster Dictionary" ;
     skos:definition "The part of a rock formation that appears at the surface of the ground."@en ;
```
geoderekh, line 147: formation has stratigraphic meaning, and the def implies that the outcrop is only where the ground surface meets a rock exposure, and that rocks above the surface of the ground are not part of the outcrop.

suggested def. A field site where bedrock is exposed at the surface of the earth, and not concealed by regolith or water. There is a high degree of confidence that samples and observations from an outcrop site relate to the bedrock at that location.

vocabularies/geological-sites.ttl
```diff
@@ -138,32 +143,69 @@ gsqst:colluvial a skos:Concept ;
 
 gsqst:outcrop a skos:Concept ;
     skos:broader gsqst:field-site ;
-    dcterms:source "Merriam-Webster Dictionary" ;
+    dct:source "Merriam-Webster Dictionary" ;
```
geoderekh, line 146: their def isnt great so i'd remove this along with the new def.

vocabularies/geological-sites.ttl
```diff
@@ -138,32 +143,69 @@ gsqst:colluvial a skos:Concept ;
...
     skos:definition "A field site from which a soil sample is taken or observation is made."@en ;
     skos:inScheme <https://linked.data.gov.au/def/geological-sites> ;
-    skos:notation "sh" ;
-    skos:prefLabel "Soil Horizon"@en .
+    skos:notation "sr" ;
+    skos:prefLabel "Saprolith"@en .
+
+gsqst:subcrop a skos:Concept ;
+    skos:altLabel "sub-outcrop"@en ;
+    skos:broader gsqst:field-site ;
```
geoderekh, line 181: should be skos:broader gsqst:outcrop ;

vocabularies/geological-sites.ttl
```diff
@@ -138,32 +143,69 @@ gsqst:colluvial a skos:Concept ;
...
+    skos:definition "The part of a rock formation under soil or sediment cover that is close to, but not outcropping at, the ground surface."@en ;
```
geoderekh, line 182: def needs to be replaced.
suggested def "A field site where bedrock is only exposed over a small area with very little topographic relief, to such an extent that the bedrock is reasonably inferred to be in-situ. There is a reasonable degree of confidence that samples and observations from a subcrop site relate to the bedrock at that location. "

vocabularies/geological-sites.ttl
KellyVance: done

vocabularies/geological-sites.ttl
KellyVance: dictionary source removed

vocabularies/geological-sites.ttl
KellyVance: done

vocabularies/geological-sites.ttl
KellyVance: done

BlakePaul:
OK.  Lets see if I can use this.  I agree with the 3 new terms and the splitting of Tailings.

However, the changing of Soil to Pedolith, does that mean that Soil will no longer be available?  I think we should retain a general "Soil" of Geological Site.

KellyVance:
> OK. Lets see if I can use this. I agree with the 3 new terms and the splitting of Tailings.
> 
> However, the changing of Soil to Pedolith, does that mean that Soil will no longer be available? I think we should retain a general "Soil" of Geological Site.

@BlakePaul Soil will be an alias of pedolith. So if you search for soil you'll still get the results. If we find in the future that we start getting lots of sites that need a greater level of distinction we can always split pedolith into its components.

PR #263, opened 2020-06-12
Multiple SKOS fixes version 2

DavidCrosswellGSQ:


KellyVance:
Notation Errors still present. Will need further fixing. But this will be sufficient to progress UAT.

PR #275, opened 2020-06-22
Update sample-relationship.ttl

KellyVance:
Have commented out hasSample so we define where samples were derived from, and just infer the inverse.
Tidied up definitions to reference where samples are derived from rather than the [w3 definitions](https://www.w3.org/TR/vocab-ssn/#Sampling-specification) so we make no claims as to the representativeness of our samples, only their provenance. 
Have added skos:inscheme to concepts to ensure they render in VocPrez.

Further work may be needed to expand this in future.

### DIFF
--- a/vocabularies/sample-relationship.ttl
+++ b/vocabularies/sample-relationship.ttl
@@ -7,40 +7,38 @@
 @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
 
 <http://linked.data.gov.au/def/sample-relationship> a owl:Ontology , skos:ConceptScheme ;
-    dcterms:created "2020-03-24T09:03:21"^^xsd:dateTime ;
+    dcterms:created "2020-03-24"^^xsd:date ;
     dcterms:creator <http://linked.data.gov.au/org/gsq> ;
-    dcterms:modified "2020-03-24T09:03:21"^^xsd:dateTime ;
+    dcterms:modified "2020-06-23"^^xsd:date ;
     dcterms:publisher <http://linked.data.gov.au/org/gsq> ;
     dcterms:source "Source information from https://www.w3.org/TR/vocab-ssn/#Sample_Relations and the SOSA Extensions ontology, modified by the Geological Survey of Queensland." ;
-    skos:definition """Relationships between Samples.
-
-This vocabulary is derived from the SOSA Ontology and the SOSA Extensions Ontology where the Samples are defined by the class sosa:Sample and the relationships are OWL objectProperties. It reuses the original ontology URIs directly and retains the relationship's non-SKSO properties too so it is a SKOS+ ontology (SKOS + other things)."""@en ;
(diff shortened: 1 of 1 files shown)


vocabularies/sample-relationship.ttl
```diff
@@ -53,15 +51,9 @@ sosa:hasOriginalSample
     skos:prefLabel "has original sample"@en ;
     sdo:domainIncludes sosa:Sample ;
     sdo:rangeIncludes sosa:Sample ;
-    skos:definition "A Link to the original sample that is related to the context sample through a chain of isSampleOf relations."@en ;
+    skos:definition "A link from the subject sample to the original sample of interest through a chain of isSampleOf relations."@en ;
```
DavidCrosswellGSQ, line 54: Could we mention "transitive" here - it is a term we are wanting people to embrace.

vocabularies/sample-relationship.ttl
```diff
@@ -7,40 +7,38 @@
...
-    dcterms:modified "2020-03-24T09:03:21"^^xsd:dateTime ;
+    dcterms:modified "2020-06-23"^^xsd:date ;
     dcterms:publisher <http://linked.data.gov.au/org/gsq> ;
     dcterms:source "Source information from https://www.w3.org/TR/vocab-ssn/#Sample_Relations and the SOSA Extensions ontology, modified by the Geological Survey of Queensland." ;
-    skos:definition """Relationships between Samples.
-
-This vocabulary is derived from the SOSA Ontology and the SOSA Extensions Ontology where the Samples are defined by the class sosa:Sample and the relationships are OWL objectProperties. It reuses the original ontology URIs directly and retains the relationship's non-SKSO properties too so it is a SKOS+ ontology (SKOS + other things)."""@en ;
+    skos:definition """Relationships between Samples. This vocabulary is derived from the SOSA Ontology and the SOSA Extensions Ontology where the Samples are defined by the class sosa:Sample and the relationships are OWL objectProperties. It reuses the original ontology URIs directly and retains the relationship's non-SKSO properties too so it is a SKOS+ ontology (SKOS + other things)."""@en ;
```
DavidCrosswellGSQ, line 15: typo non-SKSO

vocabularies/sample-relationship.ttl
KellyVance:
```suggestion
    skos:definition "A transitive link from the subject sample to the original sample of interest through a chain of isSampleOf relations between multiple successive samplings of samples"@en ;

```

Too convoluted? @DavidCrosswellGSQ

vocabularies/sample-relationship.ttl
KellyVance, line 15:
```suggestion
    skos:definition """Relationships between Samples. This vocabulary is derived from the SOSA Ontology and the SOSA Extensions Ontology where the Samples are defined by the class sosa:Sample and the relationships are OWL objectProperties. It reuses the original ontology URIs directly and retains the relationship's non-SKOS properties too so it is a SKOS+ ontology (SKOS + other things)."""@en ;

```

PR #281, opened 2020-06-25
Map collection Vocabs

CantRoger:
I can't for the life of me get http://demo.seco.tkk.fi/skosify/skosify to work...
Might need to update the guide?

Ran my ttl through here Vance, https://rdf-translator.appspot.com/ and it came out looking good, but there is stuff at the bottom that should be at the top. I have the original ttl files if you'd like me to run it through something else.


KellyVance:
Pull request cancelled. Needs of map collection met through other means

PR #285, opened 2020-07-02
Lith and Min Upload

KellyVance:
Lithology and Mineral vocabularies re-upload
Tech Check @LukeHauck and/or @DavidCrosswellGSQ 
General Review @GSQ-AI @geoderekh

No collections created yet for more specific use cases.
Given the breadth and scale of these vocabularies I suggest looking for egregious omissions rather than doing a more comprehensive check. Errors, omissions, and alterations will all become evident with usage. Deliberately omitting wider SME reviews in favour of more rapidly publishing a first draft that can be interacted with through VocPrez for a general audience and beginning a process of incremental improvement.

@DavidCrosswellGSQ both files are orders-of-magnitude larger than the other vocabularies. Any automated scripting for vocabulary refresh should omit both of these and allow them to be refreshed manually instead.


GSQ-AI:
As per the discussion with you this afternoon, @KellyVance, we should rename the vocabulary from "Lithological Observations" to "LithoType" to better describe the content and free up the name for use later.

PR #286, opened 2020-07-03
Update report detail type

KellyVance:
Update of report detail type to support PGGD ingest.
Contents may extend beyond what is required for 'balance form geoproperties' but has been included for completeness.
Suggest Address fields be concatenated into a single value for the purposes of these forms.


KellyVance:
Created from wrong branch. Will redo

PR #287, opened 2020-07-03
Update report-detail-type.ttl

KellyVance:
Update of report detail type to support PGGD ingest.
Contents may extend beyond what is required for 'balance form geoproperties' but has been included for completeness.
Suggest Address fields be concatenated into a single value for the purposes of these forms.

### DIFF
--- a/vocabularies/report-detail-type.ttl
+++ b/vocabularies/report-detail-type.ttl
@@ -10,7 +10,7 @@
         skos:ConceptScheme ;
     dcterms:created "2020-04-29"^^xsd:date ;
     dcterms:creator <http://linked.data.gov.au/org/gsq> ;
-    dcterms:modified "2020-07-03"^^xsd:date ;
+    dcterms:modified "2020-07-07"^^xsd:date ;
     dcterms:publisher <http://linked.data.gov.au/org/gsq> ;
     dcterms:source "Compiled by the Geological Survey of Queensland" ;
     skos:definition "Miscellaneous details relating to reports and documents."@en ;
@@ -22,7 +22,8 @@
             rptdt:email,
             rptdt:operating-company,
             rptdt:phone,
-            rptdt:project-name .
+            rptdt:project-name,
+            rptdt:reportable-incident .
 
 <http://linked.data.gov.au/org/gsq> a sdo:Organization ;
     sdo:name "Geological Survey of Queensland" ;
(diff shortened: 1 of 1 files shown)


KellyVance:
@LukeHauck can you please run through the usual tech checks please?

vocabularies/report-detail-type.ttl
```diff
@@ -10,26 +10,71 @@
...
-    skos:hasTopConcept rptdt:project-name,
-        rptdt:placeholder .
+    skos:hasTopConcept rptdt:address,
+            rptdt:comments,
+            rptdt:contact,
+            rptdt:contracting-company,
+            rptdt:email,
+            rptdt:operating-company,
+            rptdt:phone,
+            rptdt:project-name .
```
DavidCrosswellGSQ, line 25: Please add "Activities undertaken" as features in PGGD-04
https://rawcdn.githack.com/geological-survey-of-queensland/gsq-lodgement-portal/7137f8bd7e0720a27892fc637d24101f7147f830/pggd04.html

This could also cover "Activities to be undertaken" in PGGD-03?
https://rawcdn.githack.com/geological-survey-of-queensland/gsq-lodgement-portal/7137f8bd7e0720a27892fc637d24101f7147f830/pggd03.html

Please also add "Did any incidents occur?" (choose a better name) from PGGD-04

KellyVance: @DavidCrosswellGSQ 

Added concept _reportable-incident_ for  "Did any incidents occur?" flag.
Added altLabels to _comments_ concept to include "Activities undertaken" and "Activities to be undertaken", as reflected in the definition and generalised intent of this concept to capture planned activities and activity summaries.

DavidCrosswellGSQ: I really like the reportable-incident concept.
_Comments_, for me, is a different concept to '_Activities undertaken/to be undertaken_".
Comments would cover remarks and other unstructured data.
I think that the business will want to write a query such as "Return all activities undertaken when report type = PGGD-04".

KellyVance: Good Point. I've changed the definition for comments to be more general commentary.
Added a concept simply Labelled "Activities" so it is independent of planned or actual. so you can query
- Return all activities when report type = PGGD-04 (find actual activities), or
- Return all activities when report type = PGGD-03 or PGGD-04 (compare planned vs actual)

vocabularies/report-detail-type.ttl
```diff
@@ -35,14 +35,22 @@ rptdt:address a skos:Concept ;
     skos:prefLabel "Address"@en ;
     skos:topConceptOf <http://linked.data.gov.au/def/report-detail-type> .
 
-rptdt:comments a skos:Concept ;
-    skos:altLabel "Description"@en,
-        "Summary"@en,
+rptdt:activities a skos:Concept ;
+    skos:altLabel "Summary"@en,
```
DavidCrosswellGSQ, line 34: I like this concept now, but I would remove Summary - that belongs as an alt label in Comments

KellyVance: Moved

PR #298, opened 2020-07-21
Update geo-commodities.ttl - add ordered collection

KellyVance:
Addition of OrderedCollection to replicate hierarchical structure.
Lodgement portal form code needs to be tweaked to render this list in the defined order, and not alphabetically.

Same issue in ESDC where lodgement list is alphabetical and not as per ordered collection


KellyVance:
@DavidCrosswellGSQ I've done some digging on ordered collection and  made a syntax change. Can you please check and see if this will fix the issue of the lists currently only rendering alphabetically?

PR #301, opened 2020-07-29
geo-commodities addition of notations

KellyVance:
all concepts given notations
borrowed from existing QDEX values where possible

No content or structure change. Just adding notations. 
Therefore just need tech check @DavidCrosswellGSQ or @LukeHauck

### DIFF
--- a/vocabularies/geo-commodities.ttl
+++ b/vocabularies/geo-commodities.ttl
@@ -14,7 +14,7 @@
     dcterms:created "2020-02-17"^^xsd:date ;
     dcterms:creator <http://linked.data.gov.au/org/gsq> ;
     dcterms:publisher <http://linked.data.gov.au/org/gsq> ;
-    dcterms:modified "2020-07-15"^^xsd:date ;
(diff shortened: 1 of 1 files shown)


DavidCrosswellGSQ:
I still have issues with this vocab btw.
I think that the top concepts are actually collections. I can't see anyone entering in any of these four values into a commodity select list.
I think that there are some in the next level that are also collections - they fail the concept test.

KellyVance:
I'm comfortable that they are actually things based on the vocabulary definition and context originally put forth by CGI.
No one should be _entering_ anything by those upper concepts, but they are legitimate hierarchical groupings that would be of great interest for querying from a resource custodianship PoV.

e.g. if you were a vet you would hope no one registered an animal at your clinic as simply 'Carnivore', but it might be really useful to you to have that level of classification when you want to know how many Cats and Dogs you're treating vs, rabbits, mice, and guinea pigs. Or comparing ailment type against biological classification.

Either way, for the moment, having it in its current state doesn't break anything. So I'm happy to pause and revisit this. But it is worth having the proper philosophical discussion after we get through this stage.

PR #306, opened 2020-08-23
Added Map Series and cleaned SKOS

DavidCrosswellGSQ:
Will add these changes into the geoadmin features ontology when we get approval of this vocabulary.


KellyVance (approved):
Technically coherent as per https://rdf-translator.appspot.com/
Content seems sensible

Approved.

Greenwoodmatthew (approved):
Map Series looks good.  May need some more though about how we have a vocab with ever changing permits and constants like sub blocks and map sheets.  I may do to add the 250 sheets in as well.  Map collections refer to the 100k sheet, but other things may be senisble to add to the 250k?

DavidCrosswellGSQ:
> Map Series looks good. May need some more though about how we have a vocab with ever changing permits and constants like sub blocks and map sheets. I may do to add the 250 sheets in as well. Map collections refer to the 100k sheet, but other things may be senisble to add to the 250k?

The source of truth for permits is MyMinesOnline, so we won't keep a copy of the permits in this vocab, we just use the concept that says that a permit is this type of thing.

We can create a collection of the 100K sheets, and then another collection of the 250K sheets (or any others). 
Would the 250K maps belong to a different parent - i.e. not the Map Series parent?

Greenwoodmatthew:
> > Map Series looks good. May need some more though about how we have a vocab with ever changing permits and constants like sub blocks and map sheets. I may do to add the 250 sheets in as well. Map collections refer to the 100k sheet, but other things may be senisble to add to the 250k?
> 
> The source of truth for permits is MyMinesOnline, so we won't keep a copy of the permits in this vocab, we just use the concept that says that a permit is this type of thing.
> 
> We can create a collection of the 100K sheets, and then another collection of the 250K sheets (or any others). 
> Would the 250K maps belong to a different parent - i.e. not the Map Series parent?

@DavidCrosswellGSQ  best to ask Mark H for conventions. But I would suggest we have a 100k map series collection and a 250k map series collection as they are just references. If possible could we have each 100 sheet existing as a child of the overarching 250k sheet?

DavidCrosswellGSQ:
> If possible could we have each 100 sheet existing as a child of the overarching 250k sheet?
Yes, this is very possible.

PR #307, opened 2020-08-25
Update report-detail-type.ttl

DavidCrosswellGSQ:
Added State, Suburb, Postcode for the PGGD-03 and 04 forms.

### DIFF
--- a/vocabularies/report-detail-type.ttl
+++ b/vocabularies/report-detail-type.ttl
@@ -23,6 +23,9 @@
             rptdt:email,
             rptdt:operating-company,
             rptdt:phone,
+rptdt:state,
+rptdt:suburb,
+rptdt:postcode,
             rptdt:project-name,
             rptdt:reportable-incident .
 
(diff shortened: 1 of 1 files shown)


vocabularies/report-detail-type.ttl
```diff
@@ -23,6 +23,9 @@
             rptdt:email,
             rptdt:operating-company,
             rptdt:phone,
+rptdt:state,
+rptdt:suburb,
+rptdt:postcode,
```
KellyVance, line 28:
```suggestion
            rptdt:phone,
            rptdt:state,
            rptdt:suburb,
            rptdt:postcode,

```

Just to pretty it up.

KellyVance (approved):
Minor optional indent change suggest for aesthetics. But content and structure is good.

vocabularies/report-detail-type.ttl
DavidCrosswellGSQ: Good one - have changed the indent.

PR #310, opened 2020-08-31
Addition of Country Vocabulary

KellyVance:
Based off ISO 3166-1

Downloaded from http://dd.eionet.europa.eu/vocabulary/common/countries/view
converted rdf -> ttl using https://rdf-translator.appspot.com/
removed validation fails around multiple prefLabels per language.

Required for migration of Party data

### DIFF
--- a/vocabularies/countries.ttl
+++ b/vocabularies/countries.ttl
@@ -1,0 +1,8532 @@
+@prefix : <http://dd.eionet.europa.eu/property/> .
+@prefix adms: <http://www.w3.org/ns/adms#> .
+@prefix dcmit: <http://purl.org/dc/dcmitype/> .
+@prefix dcterms: <http://purl.org/dc/terms/> .
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
+@prefix xml: <http://www.w3.org/XML/1998/namespace> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+<http://dd.eionet.europa.eu/vocabulary/common/countries/AD> a skos:Concept ;
+    skos:exactMatch <http://dbpedia.org/resource/Andorra>,
+        <http://dd.eionet.europa.eu/vocabulary/cdda/regions/AND>,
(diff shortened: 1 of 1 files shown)


geoderekh (approved):
Approved. Datatype in the party data is not currently a vocab, and this will give us the URIs for country as per the data elements (also needed to resolve service desk request GDMPSD-142). 
@KellyVance please notify me of the link for the vocab when this branch is merged and the concepts are in vocprez.

DavidCrosswellGSQ:
@nicholascar Is there an active vocab with a SPARQL endpoint that we can use instead of creating our own?
Country doesn't appear to be in LOCI

Thanks, David

nicholascar:
DBPedia:  < http://dbpedia.org/sparql>

It’s certainly got everything you could ever want in there - everything in Wikipedia - but also a lot of junk you don’t want, e.g. see:

```
PREFIX dbo: <http://dbpedia.org/ontology/>
PREFIX dbr: <http://dbpedia.org/resource/>

SELECT ?country
WHERE {
 ?x a dbo:Country.
 ?x dbo:longName ?country.
}
ORDER BY ?country
LIMIT 100
```

I think WikiData  (<https://query.wikidata.org/>) is much higher quality so likely to have a much better list of real countries and perhaps even a way to query for current countries (as opposed to “Gaul” etc.) but can’t drum up a query for that this week as I’m out west (Winton) on holidays, so doing this on the phone.

DavidCrosswellGSQ (approved):

nicholascar:
Good to see `exactMatch` across to Geonames etc.

I’ll check with the OGC: perhaps either they have a good countries vocab or perhaps you can offload such a list to them and get a SPARQL endpoint out of it.

@rob-metalinkage, does the OGC have a countries vocab?

Future VocPrez versions will do multi-instance searching so your VP, if you keep upgrading (!) will be able to search directly across OGC vocabs without any effort.

rob-metalinkage:
OGC does not publish a countries vocabulary itself.

There are ISO and UNstats country codes. The SIRF project I ran at CSIRO loaded and cross-mapped these - but its no longer active :-( 

Country identification is a rather fraught political problem - we really need a UN solution to publishing this.  SIRF identified 29 different country vocabularies or identifier schemes present in the UN data archive :-)

geoderekh:
@KellyVance will we host this vocab on vocprez? if so, could you please let me know when you have refreshed the server so i can inform SRA of the URI to use.

KellyVance:
@DavidCrosswellGSQ my IP must have changed again and I can't access the graph to refresh the cache. Can you please oblige @geoderekh and push this update through?

PR #311, opened 2020-08-31
Added concept scheme preflabel and header info

DavidCrosswellGSQ:
Hi Vance,
I was too hasty in my previous approval of this vocab.
Have fixed the concept scheme.
Cheers, David


KellyVance (approved):
Well spotted.
Thanks for the pick-up
Approved

PR #318, opened 2020-09-18
Update gsq-roles.ttl

UbuntuScott:
Tidy up of DES roles, removing broader - narrower from data manager role

I've added a comment as a placeholder, further updates to follow conceptual discussions.


KellyVance (approved):
Changes make sense.
Passes all technical checks.
Approved!

PR #331, opened 2020-11-01
Removing Petroleum Report - Infrastructure and Petroleum Report - Cum…

talebib:
…ulative Water Production

Petroleum Report - Infrastructure, was removed from reporting requirements in 2018 under the MWOLA Act 2018.

Petroleum Report – Cumulative Water Production is no longer required. This report is lodged with the Petroleum Production Report in the LP as part of petroleum production reporting requirements under the Petroleum and Gas (General Provisions) Regulation 2017 Section 42(3)(d) and (f).

### DIFF
--- a/vocabularies/georesources-report-types.ttl
+++ b/vocabularies/georesources-report-types.ttl
@@ -419,28 +419,12 @@ grt:petroleum-report-annual a skos:Concept ;
     skos:prefLabel "Petroleum Report - Annual"@en ;
     skos:topConceptOf <http://linked.data.gov.au/def/georesource-report> .
 
-grt:petroleum-report-cumulative-water-production a skos:Concept ;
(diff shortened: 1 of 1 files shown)


KellyVance:
@talebib 

I think the intent here is to retain these report types as things that have been submitted in the past and still exist in the system, but are no longer current report types that will be accepted by the lodgement portal.

As such I would retain these as concepts, but remove them from the collections that the lodgement portal uses to populate the Report Type Selection drop-down list.

i.e. instead of removing them entirely, find the list _grt:lodgement-portal-reports a skos:Collection ;_  and just _remove grt:petroleum-report-cumulative-water-production_ and _grt:petroleum-report-infrastructure_ from that list

talebib:
Thanks Vance. Understood!

PR #332, opened 2020-11-02
Removing the reports from the LP drop-down list.

talebib:
Petroleum Report - Infrastructure was removed from reporting requirements in 2018 under the MWOLA Act 2018.

Petroleum Report - Cumulative Water Production is now lodged with the Petroleum Production Report in the LP as part of petroleum production reporting requirements under the Petroleum and Gas (General Provisions) Regulation 2017 Section 42(3)(d) and (f).


KellyVance (approved):
Change makes sense, as per email discussion of non-currency of these reports.

Approved

PR #336, opened 2020-11-06
Update ESDC - resource eng in LP

KellyVance:
Adding resource engineering into the lodgement portal collection, and changing definition to explicitly mention Production.
As per industry feedback on the lodgement portal.

### DIFF
--- a/vocabularies/earth-science-data-category.ttl
+++ b/vocabularies/earth-science-data-category.ttl
@@ -781,7 +781,9 @@ esdc:resources-engineering a skos:Concept ;
     skos:inScheme <http://linked.data.gov.au/def/earth-science-data-category> ;
     skos:notation "0914" ;
     skos:altLabel "RESOURCES ENGINEERING AND EXTRACTIVE METALLURGY"@en,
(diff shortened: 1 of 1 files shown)


vocabularies/earth-science-data-category.ttl
```diff
@@ -880,6 +882,7 @@ esdc:lodgement-form a skos:OrderedCollection ;
         esdc:engineering,
         esdc:surveying,
         esdc:drilling-engineering,
+	esdc:resources-engineering,
```
talebib, line 885: Pls consider if the 'resources' should be used in single form?

KellyVance: @talebib label as per source material https://vocabs.ardc.edu.au/repository/api/lda/anzsrc-for/resource?uri=http://purl.org/au-research/vocabulary/anzsrc-for/2008/0914

But can certainly change our label if preferred. I think there is room to add some of the narrower terms here as well

talebib (approved):
Thanks Vance.
All fine by me.

PR #342, opened 2020-11-11
File-Types from europa with collections

broughd-qld:
A set of file types from Europa (EU), with 3 collections created for use in QldGov, Some file types relate to those used by data.qld.
Have defined a new linked.data.gov.au item to use for the collections which I just came up with some placeholder text for - `@prefix qgft: <http://linked.data.gov.au/def/qg-file-type/>`

SKOS online testing identified that there are no skos:TopConcept items - @KellyVance will that be a problem? If so, I'll correct prior to reviews.


KellyVance:
It will probably work for cataloging, but I suspect that without any top concepts declared, VocPrez may not recognise and render the concepts correctly. It definitely won't harm anything to add the top concept declaration to each concept.

broughd-qld:
@KellyVance Have include the relevant topconcept, topconceptof and inscheme items. Found them in a referenced item from the vocabulary file. Have inserted the new items, not really a neat file, but it works.

KellyVance (approved):
Vocabulary meets skos checks. Content seems sound.

broughd-qld:
@DavidCrosswellGSQ  I have just removed the hasTopConcept items from the ConceptScheme. The individual file-type concepts are marked as being topConceptOf of the ConceptScheme.

I considered moving from the EU list to our own curated linked.data.gov.au list - but the EU list is well maintained and curated. If we need to add additional file types, we can add to http://linked.data.gov.au/def/qg-file-type as new concepts. Will just have to create a ConceptScheme.
cc @UbuntuScott

PR #349, opened 2020-11-26
Update geological-sites.ttl

DavidCrosswellGSQ:
Adding in Resource authority to allow the upload of Mining Leases into Geoproperties database as sites.
This will allow GSQ to create the relationship between a mine site and a resource authority (permit).


KellyVance (approved):
Easy one. Approved

PR #351, opened 2020-11-27
Create GSQ Initiatives

KellyVance:
First cut at a GSQ project/initiative vocabulary for indexing datasets against specific initiatives we may wish to group and promote together.
Very naive attempt, so may require significant feedback and modification. Please tag anyone applicable.

We may need to modify the CKAN model for datasets to accommodate this at some point, but that can be done at a later stage once this list is more definitive. Just up for general review and discussion as it isn't a critical vocabulary at this point, but the more we iterate before we need it, or want to actually implement it, the better.

### DIFF
--- a/gsq-initiatives.ttl
+++ b/gsq-initiatives.ttl
@@ -21,7 +21,7 @@
 
 gsqi:cdi a skos:Concept ;
     skos:prefLabel "Collaborative Drilling Initiative"@en ;
-    skos:definition "The Collaborative Drilling Initiative (CEI) provides industry grants to: encourage exploration drilling programs, particularly in frontier areas;  "@en ;
+    skos:definition "The Collaborative Drilling Initiative (CEI) provides industry grants to encourage exploration drilling programs, particularly in frontier areas;  "@en ;
     skos:inScheme <http://linked.data.gov.au/def/gsq-initiatives> ;
     skos:notation "CDI" ;
     skos:altLabel "CDI"@en ;
@@ -29,7 +29,7 @@ gsqi:cdi a skos:Concept ;
 
 gsqi:cei a skos:Concept ;
     skos:prefLabel "Collaborative Exploration Initiative"@en ;
-    skos:definition "The Collaborative Exploration Initiative (CEI) provides industry grants to: encourage exploration, particularly in frontier areas; support innovative exploration techniques, including machine learning; promote the discovery and development of 'new economy' minerals. "@en ;
+    skos:definition """The Collaborative Exploration Initiative (CEI) provides industry grants to: encourage exploration, particularly in frontier areas; support innovative exploration techniques, including machine learning; promote the discovery and development of 'new economy' minerals."""@en ;
     skos:inScheme <http://linked.data.gov.au/def/gsq-initiatives> ;
     skos:notation "CEI" ;
     skos:altLabel "CEI"@en ;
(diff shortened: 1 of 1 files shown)


Greenwoodmatthew (commented):
Hi Vance, this will require some re-work i think as we are crossing programs / projects under those programs etc.

I feel this would be be suited to allocate datasets / projects under the major funding initiatives we have i.e. SREP, NEMI, Future Resources, Greenfields2020.....  We could even use the nesting of vocabs collections to fit items into subset (i.e. New discoveries, vs Frontier basins under SREP)?

KellyVance:
Thanks @Greenwoodmatthew, i'll take any feedback and guidance on this to make it useful. Happy to hierarchy the concepts, create collections, or just remove the finer scale projects and roll this up to just the funding initiatives. The latter may be the most robust starting point.

mckillopm:
I notice Carbon Geostorage Initiative (CGI) is not there. This was a 4 year program 2008-2012. Also projects and Initiatives, I think there a big difference. If we are going to do every project that could include every little project we did creating GSQ records. SEEBASE is in there, which was a large project overseen by Behnam Talebi but if you include this you'd have to include the Toolebuc work, started by Melanie Fitzell and carried on by the Basin Studies group up till 2018. If you include the Toolebuc work, what about Micaela's Fort Cooper Coal Measures work which was a few years work and David Coffey's et al. Galilee Basin review? Just need some clarity on where we draw the line other wise we'll include a number of small projects which I have not mentioned.

Greenwoodmatthew:
Mike, i agree.  I think we should focus on major initiatives for the time.  NEMI, SREP, CGI, greenfields2020...  once we have these we can have nested subsets of the major initiatives.  Maybe we could get a set of these to @KellyVance once he is back.  I guess with the history of GSQ that @BlakePaul, @CourteneyD , @SalEdwards01, @mckillopm nd others have we could get a good list of the last 15-20 years of these projects.  I can set up an excel and we can have a crack?

Greenwoodmatthew:
Excel sheet with a start of funding programs.  Please fill out to help Vance with this vocab?

https://itpqld-my.sharepoint.com/:x:/g/personal/matthew_greenwood_dnrme_qld_gov_au/Ef3Dy6MeGuNCslcHsyUWUYcB6DMl6tqzPN-aH9w6zz_AeQ?e=8MzihK

broughd-qld:
@KellyVance would the initiatives and projects actually be of the type [prov:Activity](https://www.w3.org/TR/prov-o/#Activity)?

We have thought about similar things for all of the various projects and activities that we do to generate our science datasets in DES. The PROV-O model fits very well with its definition of activities, entities and agents (or Parties in GSQ terms). Activities can be associated and informed by other activities - they also have start and end times.

Maybe better to think about these in the PROV model and not squeeze into an ontology. VocPrez may need a bit of tweaking to deal with them visually - but the graphdb will be fine.

geoderekh:
Thanks for putting this one together @KellyVance . The idea is good and the content looks ok in general, though I was thinking it would be good to pull out start and end times as properties, and we will want to link the initiatives with outputs (and maybe also inputs and impacts on industry such as discoveries or tenure acquisition to track our performance). 

And on this, nice one @broughd-qld (hi by the way). I agree that PROV-O looks like a good ontology that describes (my naïve understanding of) what we are trying to do here, and I can see several additional benefits. 

Yes, I think initiatives nicely fit prov:Activity We already use prov:Agent in geoprops, and there is reasonable alignment between prov and sosa which means we could automatically generate gsqi from the initiatives/survey/activity and linkages to datasets/samples/results/entities found in geoprops. While this is currently not really needed with the number of initiatives and linked datasets, going forward it will likely grow and it would make sense to automate this to allow users to enrich our graph via crud, rather than vocab managers having one more vocab to update manually. 

Also, according to this Cox paper I found [https://www.mssanz.org.au/modsim2017/C2/cox.pdf], it makes a nice bridge to BFO and OBOE, which I see as having benefits to down the line given the critical importance of Environmental (OBOE), Social (BFO) and Governance metrics in determining the success or otherwise of mining projects and initiatives going forward. And if DES are already using it then double bonus on that front...

broughd-qld:
> And if DES are already using it then double bonus on that front...

Hi @geoderekh - we only have something basic at the moment for our metadata schema, but are thinking about we can expand and link our work on projects / activities to datasets and the products we produce. 

We have an internal project reporting system that has been in place for a number of years - which we could describe with PROV-O. Using PROV-O for projects gives us a similar language for lineage.

So, not much implemented yet but some basic PROV-O for lineage, But happy to talk and align the direction we both seem to be heading.

PS - great find with that Cox paper.

DavidCrosswellGSQ:
Hi all, as an alternative, I recommend these projects/programs/initiatives be created as Surveys in Geoprops.

Be careful not to get hung up on the definition of survey as it has been used in the past. Here is the new definition:
`The one-off event examining a geological or administrative feature. The type of survey or sampling performed to investigate a feature, site, or sample. Surveys are activities that produce samples or make one-or-many observations.`

I think that this list is instance data, not reference or classification data. Hence, it should be in our database rather than a vocab.

We can add new survey types in the survey type vocab so the surveys can be classified as Projects, Programs, QSQ Initiatives, etc..

With these GSQ initiatives in Geoprops, you get the following:
a. Add rich metadata to the Survey record (description, dates, agents, geometry, relationships between surveys)  
b. Can link to Sites and Features - with in-built PROV-O relationships.  
c. Can link to Samples, Observations and Results - with built-in PROV-O and SOSA relationships.  
d. Can link to datasets.

SalEdwards01:
Hi David

Can a survey (program or initiative) have a survey(produced seismic data and field data)?
For instance the Camooweal Seismic Project had a reconnaissance survey (field work that collected samples) in preparation for the actual seismic survey acquisition.

If so, this works for me!

And I’m not sure this email was actually meant for me. Was it Derek, Matt G, and Macca?

Cheers
Sally

DavidCrosswellGSQ:
@SalEdwards01 - that email was sent by Github to everyone in this thread.

Yes - you can create relationships between initiatives (a survey) and other surveys:
image

Greenwoodmatthew:
I feel adding the initiative as a Survey type is a massive overkill.  This is for use in KB CKAN to link to related objects, having survey within surveys (within surveys??) seems a bit like inception and creates more questions than it fixes

KellyVance:
Good discussion all. I _think_ what @DavidCrosswellGSQ is alluding to is that what we have known as a 'survey', was initially thought of as a specialisation of prov:activity, but in its truest sense is, in fact, just prov:activity. 

However, initiatives do not neatly fit the definition of a survey as we have published it, and adding a 'project' survey type would be at odds with what is already in that list. To include projects/initiatives would be stretching our definition of surveys and create some strange facets and selectable options for end users. Outside those of us here who are not deep in our discussions and developments its inclusion would be questionable.

Technically, in an ideal world, we would have Surveys and Initiatives as an independent specialised Classes of objects under the prov:activity umbrella. We could then link everything using with all the prov and sosa relationships. But what we are trying to achieve at the moment is linking related objects in the KnowledgeBase by project so future us can put on a filter/facet/query and find all the NEMI/GDMP/CEI stuff. So while it might be instance data, it will be a discrete set, and achievable through a vocabulary or ontology list similar to geofeatures. We could also use prov:startedAtTime and prov:endedAtTime to begin pulling out the start and end times of initiatives as Derek suggests.

So the options as I see them are:
1. Piggyback surveys to add initiatives but have bastardise the definition of surveys and risk overcomplicating/confusing the matter for our users.
2. Create Projects/Initiatives as their own distinct class, but by doing so create a huge piece of work to adjust the underlying Db and data models that we may not realistically be able to achieve
3. Create projects as a simple vocabulary to enable CKAN filtering, but lose the ability to have all the rich metadata associated with those initiatives.

I'm currently happily fence sitting, so welcome any-and-all thoughts.

Greenwoodmatthew:
I am leaning to the option 3.  While the ability to add descriptions and other facets of initiatives etc would be nice these should exist outside of the data model.  The vocab entry for SREP as an example could have a lot of the required into held in the def text, yes?

KellyVance:
Dead pull request. Can revisit concept after current project work concludes

PR #353, opened 2020-12-04
include definitions and not just descriptions

broughd-qld:
@UbuntuScott @DavidCrosswellGSQ Just realised that the document-licence vocabulary (ie creative commons) didn't include a skos:definition for 2 of the 3 licence types.

Basic schoolboy error. Hoping this is a super quick review.


DavidCrosswellGSQ (approved):
Oh dear, how embarrassment - I missed that totally...
FYI we haven't been using dct:description in our vocabs.

I'll approve this, but did you also want to fix the dct:creator so it was only DES?

broughd-qld:
@DavidCrosswellGSQ just fixed that. Thanks for pointing that out. And our preference is to use a definition and not a description.

DavidCrosswellGSQ (approved):

DavidCrosswellGSQ:
This vocab is now refreshed in VocPrez

PR #356, opened 2021-01-06
Update georesources-report-types.ttl

talebib:
Adding old "Collaborative Drilling Initiative - Final" as an internal report, at L1 request in order to edit old CDI reports.


DavidArmstrongGSQ (commented):
The addition to line 974 is indented via Tab. Should be consistent with spaces of other lines (not sure if actually matters or not, but better safe than sorry!). Apart from that all good.

talebib:
@DavidArmstrongGSQ the indent is not there when editing but don't know why it's indented in review screen! I don't think the indent will be an issue except that it is not consistent with other lines.

DavidArmstrongGSQ (approved):

KellyVance (approved):

PR #357, opened 2021-01-06
fix of https issues

broughd-qld:
Hi @KellyVance 
I have fixed the HTTPS issues in the 5 files for data quality and the qg agents and sites. Also updated the modified dates.
It should be good to go with a quick review.


KellyVance (approved):
Changes should address http issue.
Will refresh dependent systems in due course to test presentation of files.

PR #360, opened 2021-01-19
update file type TEST

KellyVance:
@broughd-qld I am committing this directly as a test. Will roll back as needed.

### DIFF
--- a/vocabularies/europa-file-type.ttl
+++ b/vocabularies/europa-file-type.ttl
@@ -15,6 +15,18 @@
 @prefix qgft: <http://linked.data.gov.au/def/qg-file-type/> .
 @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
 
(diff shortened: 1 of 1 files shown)


UbuntuScott:
Thanks @KellyVance 
The file type SPARQL works now 

PREFIX skos:<http://www.w3.org/2004/02/skos/core#> 
SELECT *
WHERE{
  <http://linked.data.gov.au/def/qg-file-type/des-file-types> skos:member ?concept. 
  ?concept skos:prefLabel ?prefLabel.
  OPTIONAL {?concept skos:altLabel ?altLabel}.
  FILTER(LANG(?prefLabel)="en") . 
}


Also tested successfully for package and compression types.

broughd-qld:
As does the compression and packaging types.
Thanks very much @KellyVance 
cc - @UbuntuScott

PR #366, opened 2021-02-07
Update data-access-rights.ttl

KellyVance:
Addition of _conditionally protected_ to cover resources that are protected based on specified criteria.
e.g. For reports that remain confidential as long as replacement data is submitted by the end of the resubmission period.

- [ ] content @LizDerrington 

- [x] tech @LukeHauck


LukeHauck (approved):
Tech good. Approved.

PR #370, opened 2021-02-22
Update dataset-relationship.ttl

UbuntuScott:
Updates to dataset-relationship following Nick Car review. 
+removed Owl classification 
+ corrected Prov url
+ assorted minor corrections

+ rename to dataset-relationships


broughd-qld (approved):

johnmckellar (commented):
I am not greatly familiar with all of this, but it all seems OK. My one and only question/concern is:  The namespace for all PROV-O terms is http://www.w3.org/ns/prov#.

Should not the following prefix be in the heading? 
@prefix : <http://www.w3.org/ns/prov#> .

KellyVance (approved):

UbuntuScott:
> I am not greatly familiar with all of this, but it all seems OK. My one and only question/concern is: The namespace for all PROV-O terms is http://www.w3.org/ns/prov#.
> 
> Should not the following prefix be in the heading?
> @Prefix : http://www.w3.org/ns/prov# .

Good pickup @johnmckellar 
There's a few lines where I haven't updated the prefix correctly. 
Should be referring to : 
o	prov: http://www.w3.org/ns/prov#

and not 
https://www.w3.org/TR/prov-o/.... of which there a still a few links... 
I'll submit a corrected commit in a few minutes

johnmckellar (approved):

PR #377, opened 2021-03-11
Update gsq-dataset-themes.ttl

johnmckellar:
@KellyVance, to note
Addition of prefix code/acronym.

Not sure if they are all required, as in:
 skos:member gsqdt:geoscience-data,
        gsqdt:geoscience-project-data,
        gsqdt:operational-data,
        gsqdt:physical-samples ;


KellyVance (approved):
gsqdt: is agood addition.
Consistent with other vocabularies. Makes querying and editing more definite and distinct. addition of prefix to "members" is correct.

johnmckellar:
Cheers Nick and thank you for your recent feedback.

PR #381, opened 2021-04-20
Update geological-sites.ttl

johnmckellar:
Some site types, present in the vocab, were absent from the header list. I have added them, as their absence initially threw me.


KellyVance:
@johnmckellar the header list is just the topConcepts. From what I can tell the additions you have made are all subsidiary concepts that have 'broader' terms.

johnmckellar:
OK, I'll have a closer look. Cheers

johnmckellar:
@KellyVance 
I agree. Cheers

PR #383, opened 2021-04-28
Update qgiscf definitions

broughd-qld:
Had an incorrect definition for protected in the file. Updated from the QGISCF


UbuntuScott (approved):
We weren't too careful with our reviewing safeguards here.  :fearful: 
Good pickup @broughd-qld   :+1: 

Also noting the collection is for qesd to use but other agencies can add other -cabinet or -commercial etc terms that they might need in the concept scheme.

PR #384, opened 2021-04-28
Iso dq updates

broughd-qld:
@KellyVance - updated as discussed.
@nicholascar - Have changed the complex URI to a simpler one as discussed recently.

### DIFF
--- a/vocabularies/dq-iso-19157.ttl
+++ b/vocabularies/dq-iso-19157.ttl
@@ -2,8 +2,8 @@
 # imports: http://linked.data.qld.gov.au/def/dcat
 
 @prefix dcterms: <http://purl.org/dc/terms/> .
-@prefix dq-iso-19157: <https://linked.data.gov.au/def/ISO-19157/quality-dimension/> .
-@prefix dq-iso-25012: <https://linked.data.gov.au/def/ISO-IEC-25012/quality-dimension/> .
+@prefix dq-iso-25012: <https://linked.data.gov.au/def/iso-iec-25012-data-quality-dimension/> .
+@prefix dq-iso-19157: <https://linked.data.gov.au/def/iso-19157-data-quality-dimension/> .
 @prefix dqv: <http://www.w3.org/ns/dqv#> .
 @prefix owl: <http://www.w3.org/2002/07/owl#> .
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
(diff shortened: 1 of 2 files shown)


nicholascar (approved):

nicholascar (changes requested):
The URI changes look good but make sure you do the whole validation thing in https://shacl.org/playground/. I see the dq-iso-19175.ttl file generated 26 errors - mostly (25) the same thing, sh:resultPath rdfs:isDefinedBy ; sh:resultMessage "Less than 1 values" ; so each Concept just needs rdfs:isDefinedBy <..> (the vocab URI) added. The remaining error is that the Vocab (ConceptScheme) has two `dcterms:provenance` but you're only allowed one!

broughd-qld:
@nicholascar I have made the changes after testing in schacl playground. No further errors appear.

broughd-qld:
@nicholascar For your information, when doing the testing in https://shacl.org/plaground; an error occurs when using dcterms:provenance with a language string.
From the [Dublin Core definition](https://www.dublincore.org/specifications/dublin-core/collection-description/collection-application-profile/2007-03-09/#coldctermsprovenance) it does indicate that language tags are allowable. Not sure if either I'm mis-reading the specification or a tweak to the SHACL validation is required to include language tags.

KellyVance:
@nicholascar 

> I see the dq-iso-19175.ttl file generated 26 errors - mostly (25) the same thing, sh:resultPath rdfs:isDefinedBy ; sh:resultMessage "Less than 1 values" ; so each Concept just needs rdfs:isDefinedBy <..> (the vocab URI) added. 

I suspect the majority of the vocabularies we currently have will fail that rule in their present state.
If I understand correctly is this where we would put something like the following?

dq-iso-19157:AccuracyOfATimeMeasurement
    a dqv:Dimension ; 
    a skos:Concept ;
    skos:broader dq-iso-19157:TemporalQuality ;
    skos:definition "Correctness of the temporal references of an item (reporting of error in time measurement)"@en ;
    rdfs:isDefinedBy <https://www.iso.org/standard/32575.html> ;
    skos:inScheme <https://linked.data.gov.au/def/iso-19157-data-quality-dimension> ;
    **rdfs:isDefinedBy <https://linked.data.gov.au/def/iso-19157-data-quality-dimension> ;**
    skos:prefLabel "Accuracy of a Time Measurement"@en .

nicholascar:
> If I understand correctly is this where we would put something like the following?

Yes, it's almost always just a replicate of the `skos:inScheme` property.

I thought I'd added `rdfs:isDefinedBy` too all GSQ vocabs in a PR but I can't see it (must have been some other vocab collection I updated). If you like, I can make this PR. It's a replace-in-files task using `skos:inScheme` for all Concepts in all Vocabs, so pretty easy.

KellyVance (approved):

KellyVance:
@nicholascar If you can do it easily it would be deeply appreciated. It would save me doing it as a find and replace on each file individually.

PR #386, opened 2021-05-10
updated missing terms on units vocab

broughd-qld:
There were some missing terms and types to be corrected.


broughd-qld:
Hi @KellyVance  - just a couple of minor issues which I had to correct. No super urgent rush required to update but putting on the list.

KellyVance (approved):

PR #391, opened 2021-06-08
Create DES obs properties

broughd-qld:
2 new vocabulary files to create the observed properties for DES as used in the QESD Catalogue.

1. [TERN properties](https://github.com/geological-survey-of-queensland/vocabularies/blob/f925127a89f89a83a7f3cedb76b4d29b691c025c/vocabularies/tern-obs-prob.ttl) - a version of the properties used by TERN; uses the same URIs. Likely that after review a number of the properties will be removed for our purposes, but sticking with everything as a start.
2. [Qld properties](https://github.com/geological-survey-of-queensland/vocabularies/blob/f925127a89f89a83a7f3cedb76b4d29b691c025c/vocabularies/qesd-obs-probs.ttl) - a set of additional properties used by DES; currently all water related items, that will fit under the TERN top concept of 'water properties'.

Qld properties is going to import and use the TERN properties. But unsure if I actually have that setup correctly, require a review from @KellyVance and @nicholascar  to validate that this is correct.


broughd-qld:
I should also note that the files have passed validation with the VocPub validator and the SKOS testing tool. A couple of issues with the SKOS testing tool in the Qld properties file with labels, etc related to the TopConcepts used from the TERN vocabulary.

nicholascar (changes requested):
In the _QESD Observered Properties vocabulary_:

* prefLabel is missplelled: Observered -> Observed
* top concepts have no annotations like prefLabels, e.g.:

```turtle
<http://linked.data.gov.au/def/tern-cv/123f2efe-3891-4764-9047-609c8b30e925>
   a skos:Concept ;
    skos:topConceptOf <https://linked.data.gov.au/def/qld-obsprop> ;
.
```

* You need  `rdfs:isDefinedBy` for all Concepts to indicate which ConceptScheme they originate from. So all your new terms will just be `<??> rdfs:isdefinedBy cs: ;` but the TERN terms will be `<??> rdfs:isdefinedBy <TERN-CS-URI> ;`
    * I can't tell if you've actually added new terms or if you've just re-id'd existing tern ones, e.g.:

```turtle
<https://linked.data.gov.au/def/qld-obsprop/daad4aac-0a34-5715-8a84-38cbe540b16e> 
    a skos:Concept ; 
    skos:inScheme <https://linked.data.gov.au/def/qld-obsprop> ; 
    skos:prefLabel "Fluoride as F"@en ; 
    skos:definition "Fluoride as F"@en ; 
    skos:broader <https://linked.data.gov.au/def/qld-obsprop/50fe54ff-7559-5f90-9769-bd40511582f6> ; 
    qudt:hasQuantityKind quantitykind:MassDensity ; 
    dcterms:created "2021-06-07"^^xsd:date ; 
    dcterms:modified "2021-06-07"^^xsd:date ;
.
```

The example Concept above has a _qld-obsprop_ URI but isn't it copied out of a TERN vocab? There are two types of Concept URIs in this vocab: 

1. <https://linked.data.gov.au/def/qld-obsprop/daad4aac-0a34-5715-8a84-38cbe540b16e> 
2. <http://linked.data.gov.au/def/tern-cv/123f2efe-3891-4764-9047-609c8b30e925>

And I don't really know why. If you are just profiling the TERN vocab, just stick to their URIs, so the only new URI you will need - the URI you are registering - will be for the ConceptScheme, not individual Concepts

broughd-qld:
@nicholascar I have made the updates based on your review.

In terms of the URIs, where there is a TERN uri I am using it and new terms have a new URI that we are creating in a concept scheme. Both of the files. The intention is that both sets of concepts (TERN and DES) would be combined into a single vocabulary list. TERN doesn't include all of the observed properties that we will need.

I was trying to keep a degree of separation between the TERN and DES concepts to allow for easier management and updating. For the time being we can use the top level concepts from TERN without having to create additional concepts.

Does using owl:imports for the TERN vocab in the DES one bring them together?

nicholascar:
> where there is a TERN uri I am using it and new terms have a new URI 

Great, the best approach! I just couldn't be sure of that.

> Does using `owl:imports` for the TERN vocab in the DES one bring them together? Or do I need to combine this into a single set of terms?

In theory it does but `owl:imports` is tool-dependent, i.e. some copy ' paste mechanism must be implemented in a tool for it to work, and VocPRez can't do this. If you `owl:imports` an external vocab to one in VP, VP would have to resolve that vocab from source, bring it into that vocab's graph and serve it's content up. It _could_ be done but it would require a fair few now functions not yet present.

I suggest you just manually copy 'n paste all the Concepts you wish to appear in your vocab unless they are just "hand off" links like your vocab's top concepts point to other Concepts elsewhere but you don't wish those concepts to appear in your vocab directly. If this is the case, use `skos:broadMatch` [1] rather than `skos:broader`. The mapping properties are there for these sorts of hand-offs but often you want to incorporate other vocab's Concepts directly in your vocab, as you've done for the majority of them here.

[1] https://www.w3.org/TR/skos-reference/#mapping

broughd-qld:
> I suggest you just manually copy 'n paste all the Concepts you wish to appear in your vocab

Right - I will combine the 2 individual vocabularies into a single file. Please hold off on any further reviews while I do the consolidation.

broughd-qld:
@nicholascar @UbuntuScott @KellyVance - I have consolidated the concepts into a single vocabulary and file. Which is now ready for review.

@KellyVance  - I have left the file that I used from TERN. Probably don't need it loaded VocPrez, but will be a good reference as we move forward. What is your workflow for those types of items?

UbuntuScott (changes requested):
In the qesd-obs-probs .ttl
1. 
Should position x (and y) coordinate have broader term of      skos:broader <http://linked.data.gov.au/def/tern-cv/9e03a73d-8e89-4187-9f96-0306aad93458> ;  
= physical parameter?


2.  should [water count](http://linked.data.gov.au/def/tern-cv/2cf66d7b-6625-44d5-85d9-91f350afe7f0) and water prevalence be given a new broader term under "water parameter" or moved to "l[and surface](http://linked.data.gov.au/def/tern-cv/6f726a59-c330-44eb-898e-7d42d5fbdada)"?

3. All the water parameter 'headings' also need to be skos concepts 
e.g. 
<https://linked.data.gov.au/def/qld-obsprop/50fe54ff-7559-5f90-9769-bd40511582f6>
added this ------>     a skos:Concept ;
    skos:prefLabel "Cations, Anions, Alkalinity & Hardness"@en ;	
    skos:altLabel "Cations, Anions, Alkalinity & Hardness-WQI Grouping"@en ;	
    skos:definition "A grouping of properties for Cations, Anions, Alkalinity & Hardness."@en ;	
...
....

broughd-qld:
> 1.
> Should position x (and y) coordinate have broader term of skos:broader http://linked.data.gov.au/def/tern-cv/9e03a73d-8e89-4187-9f96-0306aad93458 ;
> = physical parameter?

@UbuntuScott - I have left as is given this is directly from TERN.

> 2. should water count and water prevalence be given a new broader term under "water parameter" or moved to "land surface"?

Left this as is since directly from TERN.

> 3. All the water parameter 'headings' also need to be skos concepts

Updated - good pick up.

broughd-qld:
Additional validation reviews and validation, including visual exploration of the graph identified some deficiencies which have since been corrected.

UbuntuScott (approved):

vocabularies/qesd-obs-props.ttl
```diff
@@ -0,0 +1,8937 @@
...
+        <http://linked.data.gov.au/def/tern-cv/777773d0-cfb4-44ff-85e2-08b48e10ac52>,
+        #<http://linked.data.gov.au/def/tern-cv/8e7a785a-8c57-4415-855f-cb6d35ad58bb>,
+        <http://linked.data.gov.au/def/tern-cv/8fa1c4bd-73b3-432a-ab49-9ee89928e6d9>,
+        <http://linked.data.gov.au/def/tern-cv/99bd0128-54d4-4a2b-8a2c-48010eb07d26>,
+        <http://linked.data.gov.au/def/tern-cv/9e03a73d-8e89-4187-9f96-0306aad93458>,
+        <http://linked.data.gov.au/def/tern-cv/bb5c1354-8d70-47b2-b333-87e3bf1157d6>,
+        <http://linked.data.gov.au/def/tern-cv/c_012ebf32>,
+        <http://linked.data.gov.au/def/tern-cv/d62a4664-89a1-49a1-9ef0-f30ee2c6d22b> ;
+    skos:narrower <http://linked.data.gov.au/def/tern-cv/0986ce8c-4a8c-4132-91f9-b7380d45ce73>,
+        <http://linked.data.gov.au/def/tern-cv/87ed6ce4-3354-4a3f-ba15-2b7cb0caf3da> ;
```
KellyVance, line 38: Q: How is skos:narrower different to skos:hasTopConcept for the concept scheme?
Should the former not appear in the list of top concepts for the vocabulary?

broughd-qld: I have updated. Was following the pattern from TERN - but since we have our own ConceptScheme we can make this tweak.

KellyVance (approved):

PR #393, opened 2021-06-28
update result - addition of specific angle types

KellyVance:
Set-up for specific recording of dip, inclination, and azimuth as required for boreholes.
For if/when we migrate any data beyond the simple at-surface borehole azimuth and inclination into geoproperties.

### DIFF
--- a/vocabularies/result-type.ttl
+++ b/vocabularies/result-type.ttl
@@ -111,6 +111,30 @@ rslt:angle a skos:Concept ;
     skos:prefLabel "Angle"@en ;
     skos:topConceptOf <https://linked.data.gov.au/def/result-type> .
 
+rslt:azimuth a skos:Concept ;
+    skos:definition "An angle in the horizontal plane relative to North. By default azimuth is relative to True North, but may also be qualified as relative to Magnetic North or Grid North"@en ;
+    rdfs:isDefinedBy <https://linked.data.gov.au/def/result-type> ;
+    skos:inScheme <https://linked.data.gov.au/def/result-type> ;
+    skos:notation "razi" ;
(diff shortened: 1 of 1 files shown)


LukeHauck (approved):
Tech good. Approved.

geoderekh (commented):
content looks good, though i question the logic and business processes for what we are storing as result. 
Boreholes have azi and dip attributes, and we could easily transform all of the inclinations where purpose=petrol to dip notation ##inclination=-1*(dip-90)## 
So what is the use case for putting these in result? The only reason i can think of is if we want to store borehole surveys using survey/obs/result - but im not sure if this is a deviation from an existing industry standard. In any case, we need to clarify when to use borehole.azimuth/borehole.dip vs result_type.

KellyVance:
@geoderekh 
This is more a set up so we can, rather than a commitment that we will.
As much for ongoing expansion and completeness of the vocabulary for broad usage, as it is to enable GSQ specific usage.
Having the terms defined so users know what type of value they are supplying or consuming is valuable regardless of how our database stores those values, or at what level-of-detail.

The old DB borehole orientation table has surface and subsurface measurements, the decision whether we do migrate those subsurface recordings as a legacy data holding (separate from ongoing modern industry standard practice) is a business decision, but this change enables the ability to do that should we so choose.
This distinction also ensures we can inform industry and internal users of the difference between Dip and Inclination as we define them, lest they be conflated or confused. We will store values as dip or inclination _only_ in the Db, but we may create views using a transformation to suit the different industry groups, and so we should be able to clearly articulate what those values mean without assuming all users are aware of the different measurement types.

PR #394, opened 2021-06-30
Fix qg-data-licenses URI errors

broughd-qld:
Changes to the URIs were not completed properly in migrating the schema name.

Simple find and replace done.


UbuntuScott (approved):
a very simple correction fix Dan
Approved!

PR #395, opened 2021-07-12
Removed dummy definitions in sample-detail-type.

DavidArmstrongGSQ:
My first Vocab! So be nice. 

Added in four sample-detail types relating to cores, "Pallet weight, pallet condition, core condition and core tray condition". Definitions might be a bit vague, so feel free to change.


johnmckellar (approved):

johnmckellar:
I was the only reviewer, and the commit has been merged by David in to the Master. So, I take it from everyone that they will be happy if I delete the branch?

PR #400, opened 2021-09-05
Site resource update

KellyVance:
Changing "Rehabilitation" from a resource status to a site status, as represents real world application, philosophical coherence, and understanding of key stakeholders such e.g. Hubs, RSHQ,


DavidArmstrongGSQ (commented):
Approved from my end

LukeHauck (approved):
Tech good. Approved.

PR #405, opened 2021-12-13
Added three new georesources report types

DavidArmstrongGSQ:
Three new georesource report types  for coal quarterly statistics, mineral annual and extractives annual.


talebib (approved):
All good. Thanks.

PR #406, opened 2021-12-13
Removed RPI reports from production, updated UAT to correct structure

DavidArmstrongGSQ:


talebib (approved):
Thanks David. All good!
Overlooked underscore previously!

PR #407, opened 2021-12-14
Fixed type, coal-quarterly-statistics. 

DavidArmstrongGSQ:
Added prod report types to two collections, internal lodgement portal and lodgement portal reports


katewd (approved):
go for it

PR #412, opened 2022-01-20
Update site-status.ttl

KellyVance:
add production status collection and reword active status


DavidArmstrongGSQ (commented):
Looks good too me. Approved

PR #420, opened 2022-02-17
Update site-relationships.ttl

KellyVance:
adding project and temporal relationships for spatially bound entities (sites)

@LukeHauck please check with validators
@katewd please check that additions meet our needs of relating mines to mine complexes and projects.

### DIFF
--- a/uat-vocabularies/site-relationships.ttl
+++ b/uat-vocabularies/site-relationships.ttl
@@ -21,12 +21,14 @@
         strel:boreholeIntersect,
         strel:contains,
         strel:covers,
+        strel:crosses,
         strel:disjoint,
         strel:equals,
         strel:injectionExtractionPair,
         strel:hasPart,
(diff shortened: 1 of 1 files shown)


LukeHauck (approved):
Tech good. Approved.

katewd (changes requested):
I think the definitions on lines 191 and 199 (hasPart and partOf) should be swapped. Also, additions to line 251 doesn't make sense to me. Can you make this clearer?

KellyVance:
> I think the definitions on lines 191 and 199 (hasPart and partOf) should be swapped. Also, additions to line 251 doesn't make sense to me. Can you make this clearer?

Edits made, but not sure if it is any clearer. Have included link in Concept scheme definition to direct to some visual representations of the spatial relationships.

katewd (approved):
Looks good to me

PR #421, opened 2022-02-17
Update site-detail-type.ttl

KellyVance:
The flag thing for reporting on multi-site projects


katewd (approved):
Thanks for that Vance. Approved.

PR #425, opened 2022-03-09

johnmckellar:
Have updated geof, primarily to accord with rur vocab:
1. Lithodemic Feature deleted
2. Complex Rock Unit, Intrusive Rock Unit, Tectonometamorphic Rock Unit introduced
3. Tectonostratigraphic Entity changed to Tectonostratigraphic Rock Unit, with broader Tectonic Feature removed
4. Tectonic Entity deleted and merged with Tectonic Feature
5. Some rock ranks removed: suite, supersuite, etc. (from memory)
6. Igneous association 'revised'.

### DIFF
--- a/vocabularies/geologic-feature-types.ttl
+++ b/vocabularies/geologic-feature-types.ttl
@@ -120,20 +120,29 @@ geof:ChronostratigraphicUnit a skos:Concept ;
         <https://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.605.7824>,
         <https://doi.org/10.1306/05230505015> ;
     skos:broader geof:StratigraphicUnit ;
-    skos:definition "As the term implies, Chronostratigraphic Units are Time-Rock Units. They are bodies of rocks, layered or unlayered, that were formed during a specified interval of geologic time. [The units of geologic time during which chronostratigraphic units were formed are called geochronologic (geological-time) units.] (ICS, Online, December, 2019). NB: The North American Stratigraphic Code (2005) and Owen (2009) advise that geochronologic units should only be associated with Lithodemic (non-stratigraphic) Units."@en ;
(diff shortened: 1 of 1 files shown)


johnmckellar:
There is a typo for solidifies. I'll fix this later.

nicholascar:
@johnmckellar you need to put the file through the VocPub validator at http://rdftools.surroundaustralia.com/validate. When I do this I get lots of errors but they are all the same type which is:

"You _MUST NOT_ use dcterms:source property more than once on any object."

The issue here is you're interpreting the property `dcterms:source` slightly differential to how the vocabulary profile intends it four use. I think you mean `dcterms:references`, so just convert `dcterms:source` to that. 

The VocPub profile reserves `dcterms:source` for a data source that's been reinterpreted to the Concept, so that's a little different to an academic or informative source, which is what `dcterms:references` should be used for. Due to `dcterms:source` being a data source, you can see why only one per Concept is allowed.

I switched `dcterms:source` for `dcterms:references` in Rock Unit Rank similarly (see https://github.com/geological-survey-of-queensland/vocabularies/blob/master/uat-vocabularies/rock-unit-rank.ttl).

johnmckellar:
Thanks Nick.

I did not forget your comments on rur not to use dcterms:source more than once, but did not fully comprehend about using it only once per concept and for it being reserved for a data source, as opposed to an academic source.

I ran the file through your rdf converter and corrected errors, but I did not realise that I also needed to run it through your validator.

Cheers,

John

johnmckellar:
When I go through the validation process, and I did try to do this before I uploaded the file, it doesn't matter which way I go, either pasting the file into the validator (Input Document) or uploading it thereto, I get the same result when I hit the validate button on the top RHS: NIL.

So, what am I doing incorrectly or not doing? I corrected the file with your rdf converter, there only being a few errors identified with this, but, as stated, I got nowhere with the validator.

KellyVance (approved):
Defer detailed content review by @predam1 and @BlakePaul 
However, technically seems to be sound, and at a high level the content is logically structured and well fomed.

KellyVance:
For this iteration or a future, can we please do our first expansion of GeoResourceAccumulation to include this hierarchy
- GeoResourceAccumulation
    - Petroleum Accumulation
        - Prospect
        - Play
        - Petroleum Field
        - Petroleum Reservoir Unit

johnmckellar:
Sure Vance, I’ll add it to the list ☺

johnmckellar:
Vance has reviewed this, and without waiting for other reviewers, I am merging this PR with the main.

PR #432, opened 2022-04-08
SKOS version of Timescale

johnmckellar:
Vance, This vocab has been sitting on the backburner for a while ... and it's about time that it saw the light of day, as we need it as a dropdown for rock-unit ages. There are headers issues that need to be sorted, and, Nick, there is an issue deserializing the source. I have played around with it further today, but it has got to the point where it becomes time-wasting. Cheers, John. 

PS: Vance, This is another reason why we need Nick onboard!!

### DIFF
--- a/geologic-time-scale-skos.ttl
+++ b/geologic-time-scale-skos.ttl
@@ -1,755 +1,6712 @@
-@prefix dcterms: <http://purl.org/dc/terms/> .
-@prefix owl: <http://www.w3.org/2002/07/owl#> .
-@prefix gts: <http://resource.geosciml.org/ontology/timescale/gts#> .
-@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
-@prefix time: <http://www.w3.org/2006/time#> .
-@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
-@prefix ischart: <http://resource.geosciml.org/classifier/ics/ischart/> .
-@prefix rur: <https://linked.data.gov.au/def/geologic-time-scale/> .
-@prefix sdo: <https://schema.org/> .
+PREFIX cs: <http://resource.geosciml.org/vocabulary/timescale/skos-demo>
+PREFIX dcterms: <http://purl.org/dc/terms/>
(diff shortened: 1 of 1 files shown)


KellyVance:
Content generally looks good. But I wonder if dct:isPartOf is a more accurate description than skos:broader
@nicholascar 

Remember that for our priorties of standing up the rock units our first pass _only_ needs the name, type, and status. These are our priorities to get up to a useable standard. Ages and all the other interesting things can only be associated with the rock units once they exist in the first place.

nicholascar:
`dct:isPartOf` would work IFF the model and display system were predicated on it! But they - SKOS / VocPub and VocPrez are predicated on SKOS, so we have to go with SKOS terms.

The real answer is to use OWL TIME temporal relations, so not either `dct:isPartOf` or `skos:broader` etc., but VocPrez doesn't know about those enough to make nice navigable UI elements like Concept Hierarchies.

nicholascar:
I have actually produced a much enhanced SKOS version of the Timescale: https://github.com/i-c-stratigraphy/geologic-timescale-kg/blob/master/rdf/taxonomies/gts-skos.ttl

You can see it displayed in the test ICS VocPrez over here: http://vocabs.stratigraphy.org/vocab/chart

Can you please replace this file with that one? You can just copy it out of GitHub and it will validate fine for VP.

When the ICS copy is up and running for real, you will be able to deprecate any copies you have in favour of that.

johnmckellar:
Thanks for the comment, Nick

johnmckellar:
Thanks again Nick. Will do. Validation was an issue.👍

nicholascar (approved):

KellyVance:
For reference, I have uploaded to our VocPrez UAT for a readable interface for review https://vocabs.uat.gsq.digital/object?uri=http://resource.geosciml.org/vocabulary/timescale/skos-demo

KellyVance (approved):

johnmckellar:
Seeing that this PR has been approved by Nick and Vance, I have merged this branch with the Master.

@BlakePaul , @predam1, I will send you a copy as a Word file.

PR #434, opened 2022-04-18
site update for LGAs and coincident geofeatures

KellyVance:
to enable report-site update with appropriate relationships
https://vocabs.uat.gsq.digital/object?uri=http://linked.data.gov.au/def/geological-sites


LukeHauck (approved):
Tech and Content good. Approved.

PR #438, opened 2022-06-09

johnmckellar:
Hi Paul, Vance, Nick,

For your review, I have added primary regolith-landform mapping units (RLUs) to geologic-feature-types. There are some secondary regolith elements in there, but there is a long list of them. I have **separately** dealt with about two-thirds of them (from the CSIRO CRC LEME/GSWA regolith-landform classification), but many of them lack definition, and it was slow going to define them. So, I will add them at a later time. Nonetheless, the regolith elements associated with any of the coded RLUs are listed under each.

We may have to reconsider these relationships:  skos:related geof:DuricrustResidualRelict ;  skos:related geof:DuricrustResidualRelict ;

Cheers,
John


KellyVance (approved):
Meets technical specification
As best I can tell content makes sense

PR #439, opened 2022-06-29
Update dataset-relationships.ttl

KellyVance:
@broughd-qld removing duplicate concepts due to slash in URI
have opted to change top concept list to be consistent with concept blocks, rather than vice versa. Please advise/review


broughd-qld:
Hi @KellyVance - I'm doing to checking on this. We have custom logic in our catalogue that handles relationships - specifically to keep the relationships and inverse relationship in check between datasets.

It's just taking a bit of time to dig through.

broughd-qld (approved):
These changes are fine and only impact on a single term in the DES catalogue. Good pickup @KellyVance on the issue.

PR #440, opened 2022-07-26
Update gsq-roles.ttl

KellyVance:
For a quick turn around. 
GSQ only vocabulary so can get 'good enough' for immediate use and revise definition once active.


CantRoger (approved):

Greenwoodmatthew (commented):
Looks good to me

PR #442, opened 2022-08-09
Establish UAT branch and Restructure Vocabulary Repo

KellyVance:
**This review is not urgent**

Establishing standing UAT branch.
In future all developments should be merged into UAT branch for testing in [VocPrez UAT](https://vocabs.uat.gsq.digital/)

Restructure of SubFolders
vocabularies-gsq - Administered and used by GSQ
vocabularies-des - Administered and used by DES
vocabulaires-qldgov - Administered by DECO and able to be used through all of government

eliminating uat-vocabularies folder in favour of a uat branch (i.e. how GitHub should be used)

De-duplicated files. Should be one-and-only-one copy of each vocabulary in the vocabulary repo, with admin and custodianship assigned to one group.


katewd (approved):
Better use of git branching and makes vocab responsibilities clearer. The only file I think needs work is the 'Feature Status' TTL, as the definitions don't differentiate terms clearly and may not all be needed. Please check that file. Otherwise, the restructure is good. Approve.

KellyVance:
> Better use of git branching and makes vocab responsibilities clearer. The only file I think needs work is the 'Feature Status' TTL, as the definitions don't differentiate terms clearly and may not all be needed. Please check that file. Otherwise, the restructure is good. Approve.

feature status has now been synced with recent update to clarify definitions.

KellyVance:
Pushing through restructure. All files are maintained. Only folder structure is changing

PR #446, opened 2022-09-15
Update site-relationships.ttl

johnmckellar:
Hi Vance, Paul,
Please review my additions to the site-relationship vocab, for geologic-feature relationships, presently for those associated mainly with rock units. Vance,
You may not like my compounded terminology, and you are welcome to suggest alternatives. But, in condensing each of them, it becomes more difficult to differentiate them from their inverse. Note that I have considered how they would appear in an alphabetical drop-down, so that similar terms are listed together in order (and not spread out all over the place). Once we have decided on the terminology, I can add the additions to the objectList, without which the amended file cannot be validated. Paul,
Please consider the terms and definitions provided, to suggest deletions, amendments and omissions.

Also, for both of you, please let  me know, if I have gone 'off-track'!!


KellyVance:
Closing this while we approve LPX+ update, to resubmit with requested changes as discussed after approval of other PR

PR #451, opened 2022-10-04
Update earth-science-data-category.ttl

geoderekh:
What: Addition of new top concept - "information and computing sciences" - taken from ANZSRC2020. Narrower terms can be added if needed later. 

Why: needed for tagging CEI and other reports using novel methods. Need also expected to grow because of increased use of machine learning tools. 

needs tech check and validation before merge
thanks


johnmckellar (approved):
Validates OK, Derek, and aligns with ANZSRC

LukeHauck (approved):
Tech good. Approved.

PR #459, opened 2023-02-05

johnmckellar:
@nicholascar @predam1 @mckillopm @LizDerrington @CourteneyD @vladrizm @Steven-Bowden 

I Have added the upper level framework for concepts under georesource accumulation ... and, Andrew, I have also added the terms that you required for petroleum. Placement of resource accumulation, however, remains open to discussion, whether in geologic-feature types or separate from this ontology. This work was undertaken in last November, before I went on leave.

Potentially, it is up to Derek, Vlad & Courtney, to develop what is needed for the ontology for Minerals; Micaela & Mike, for coal; and Andrew & Liz, for petroleum. Suggestions are welcome here, as above for placement. The separate ontologies for each resource type need to have a common theme and a purpose that meets our/your needs.

### DIFF
--- a/vocabularies-gsq/geologic-feature-types.ttl
+++ b/vocabularies-gsq/geologic-feature-types.ttl
@@ -6,38 +6,8 @@
 @prefix skos: <http://www.w3.org/2004/02/skos/core#> .
 @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
 @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
-
-<https://linked.data.gov.au/def/geofeatures> a owl:NamedIndividual,
-        owl:Ontology,
-        skos:ConceptScheme,
-        prof:Profile ;
-    dcterms:contributor [ a owl:NamedIndividual,
-                sdo:Person ;
-            sdo:affiliation <https://linked.data.gov.au/org/gsq> ;
-            sdo:email <mailto:derek.hoy@dnrme.qld.gov.au> ;
-            sdo:identifier <mailto:derek.hoy@dnrme.qld.gov.au> ;
-            sdo:name "Derek Hoy" ],
-        <https://linked.data.gov.au/org/gsq> ;
-    dcterms:created "2020-01-23"^^xsd:date ;
-    dcterms:creator <https://orcid.org/0000-0001-5489-9590>,
-        <https://orcid.org/0000-0002-8742-7730> ;
-    dcterms:modified "2022-09-28"^^xsd:date ;
-    dcterms:publisher <https://linked.data.gov.au/org/gsq> ;
-    owl:imports <http://www.opengis.net/ont/geosparql>,
-        <http://www.w3.org/2004/02/skos/core> ;
-    owl:versionIRI <https://linked.data.gov.au/def/geofeatures/0.5> ;
-    owl:versionInfo "Most ontology content but not complete diagrammatic coverage." ;
-    skos:definition "This vocabulary lists types of geological features relevant to the duties of the Geological Survey of Queensland. It is automatically derived from GSQ's Geologic Features Ontology (https://linked.data.gov.au/def/geofeatures)"@en ;
-    skos:hasTopConcept <http://sweetontology.net/realmGeol/GeologicFeature> ;
-    skos:prefLabel "Geologic Feature Types"@en ;
-    prof:isProfileOf <http://sweetontology.net/realmGeol> ;
-    sdo:codeRepository <https://github.com/geological-survey-of-queensland/geologic-features-ont> .
-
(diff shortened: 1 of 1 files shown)


geoderekh (changes requested):
As we discussed today, I like the addition of the high-level terms, and support the additonal terms in principal, but complained about the semantics as I am uncomfortable with the term "resource". I want to avoid a situation where language becomes entrenched he underlying definitions that include references to economic value, economic quantities, etc. The definition of what is economic changes regularly. Another reason is that "resource" is a defined term under the JORC code that mixes geological with economic and confidence factors (see https://jorc.org/docs/JORC_code_2012.pdf). JORC is focussed on standardising the terminology for reporting of sampling and results related to the underlying geological features on the basis of geological knowledge and confidence, as well as modifying factors (consideration of mining, processing, metallurgical, infrastructure, economic, marketing, legal, environment, social and government factors). JORC is also restricted to the domain of ASX-listed exploration/mining industry, though it is similar to other standards for companies that trade publically on international stock exchanges. It does not, however, apply to privately listed companies which exist here in QLD, and by reusing jorc:resource these mineral deposits would need to be pushed into a scheme they are not strictly compliant with, even though a "resource" and a "jorc:resource" are similar classes of things. 
As this vocab deals with the classes of things that are geological features, I propose that the classes therein are restricted to features that are defined by their geological (including physical, spatial, and temporal) properties. The minerals team (more than just me) should provide input to ensure that our definitions are well thought our and fit for purpose. 
To have a hack at it, a mineral example would be geof:mineral-deposit/mineral-accumulation, which might be roughly defined something like a contiguous or near contiguous body with anomalously high concentration of specific minerals or groups of minerals. e.g. a copper mineral deposit is a large and contiguous body of rock that contains a high relative abundance of copper-bearing minerals; the anomalous geoprop:bulk-geochemical-composition is indicated by a spatial cluster of results with a high copper concentration. It is the body of rocks with a specific set of geological properties that are of interest. A mineral deposit is a larger feature than a mineral occurrence.  A "resource" relates to estimations made at various confidence intervals (inferred, indicated, measured) of some geological properties of the mineral deposit namely grade (concentration) and amount (tonnage) with statistical likelihood terms defining the confidence-categories. 
Often the first dealing we have with a mineral deposit is when a 'maiden inferred resource' is calculated, but the geological feature we are considering existed for many millions of years prior. The event and mineral system that altered the geological properties of some rocks (that concentrated certain minerals/elements in one area of the crust) to the extent that the rocks become a mineral deposit are of interest to us so the ontology needs to be rooted in quantifiable geological results and properties. 
Another hack from me: the term Petroleum Reservoir as a narrower term of reservoir seems to make sense to me (as a non-petroleum person) as I interpret the definition to mean something like geological feature with high geoprop:porosity where pore fluid has high geoprop:petroleum-concentration. Making this vocabulary/ontology work with higher-order logic will require us to create restrictions on feature properties by formalising the definitions using OWL instead of plain text definitions, which Nick can certainly help us with. 

In addition to the semantics discussed above, I also question the logic behind geof:GeoResourceAssociatedGeoFeature since this is not a feature type at all but the object in a triple with (geof:georesource-feature featrel:generic-relationship geof:GeoResourceAssociatedGeoFeature). Using this as a feature type would lead to us losing information about the true (geological) feature type of the object feature because each feature must have exactly 1 type. For the use case presented here, we will be able to infer which features are related by simply looking for the object in the above triple (this is in the geoprops ontology/data model). 
In summation, I generally support the addition of high-level terms but identified a semantic issue with the usage of "mineral resource" which may not apply to all of our data because it is a prescribed JORC term. Mineral deposit may be more appropriate, but the Minerals team (and likewise other subject matter expert groups) must be allowed to provide input to the terms that intersect with their domain of specialty before we approve this PR.

johnmckellar:
Thanks for the feedback Derek. As there is an issue using 'resource', and, if we just omit the term, we are left with:

geof:GeoAccumulation a skos:Concept ;
geof:MineralsAndCoalAccumulation a skos:Concept ; <<< Mineral/Minerals??
geof:MineralsAccumulation a skos:Concept ; <<< Mineral/Minerals??
geof:CoalAccumulation a skos:Concept ;
geof:PetroleumAccumulation a skos:Concept ;
geof:ConventionalPetroleumAccumulation a skos:Concept ;
geof:UnconventionalPetroleumAccumulation a skos:Concept ;
geof:GeothermalEnergyAccumulation a skos:Concept ; 
geof:QuarriesAndExtractiveAccumulation a skos:Concept ;
geof:GroundwaterAccumulation a skos:Concept ;

Is this an acceptable alternative? However, does it apply equally well to all of the above concepts?

nicholascar:
I have fixed the conflicts here and ensured the file is technically valid. I haven't altered content.

I think this vocab should be passed but issued with an experimental status as it will be much easier to review further in the VocPrez hierarchy.

PR #460, opened 2023-02-23
Update geo-commodities.ttl

GSQ-AI:
Deleting cdt:phosphorus from the lodgement-form collection, as it is a hanging term with no actual definition and causing the vocprez instance to crash when opening it and the collection.


LukeHauck (approved):
Change makes sense and tech is all good. Approved.

Steven-Bowden (approved):
Looks fine to me.

PR #461, opened 2023-02-23
Update site-status.ttl

johnmckellar:
@nicholascar @LizDerrington @GSQ-AI 
I have made a few very minor grammatical changes, which can be approved without debate. But, my real intention here is to flag the issue raised yesterday by Nick, re: 'never used' vs 'never/not drilled'. We have the former in the vocab, intending to mean the latter. So, a little further down the track, we need to change this concept, as a borehole can be drilled, but never used for whatever reason.

### DIFF
--- a/vocabularies-gsq/site-status.ttl
+++ b/vocabularies-gsq/site-status.ttl
@@ -12,7 +12,7 @@
     dcterms:modified "2022-05-12"^^xsd:date ;
     dcterms:publisher <http://linked.data.gov.au/org/gsq> ;
     dcterms:provenance "Compiled by the Geological Survey of Queensland. Mapped to GeoSciML Mine Status and amalgamation with Borehole status." ;
-    skos:definition "The status of an individual site (e.g. mine, well) or single activity (e.g.survey) describing the current state within the lifecycle from proposed through operational through to abandonment. Lifecycle stages are commonly related to regulatory conditions administered by government agencies."@en ;
+    skos:definition "The status of an individual site (e.g., mine, well) or single activity (e.g., survey) describing the current state within the lifecycle from proposed through operational through to abandonment. Lifecycle stages are commonly related to regulatory conditions administered by government agencies."@en ;
     skos:scopeNote "Site status describes the lifecycle of a single discrete activity, Whereas the resource project lifecycle describes the collective lifecycle status of a body of work that typically consists of multiple activities and sites"@en ;
 
     skos:hasTopConcept ststs:abandoned,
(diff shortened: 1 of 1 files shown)


GSQ-AI:
The term "Never Used" is referring to the site, not the borehole. The "Site" is _used_ if an activity has been undertaken on it, even if that activity does not reach completion for any reason. The site is 'never used' if we create the site as a part of a proposal (PGGD-01 form, application for tenure via MMOL, or as a part of a project), and then for some reason, the proposed activities are then cancelled. The site becomes "Never Used", because no activity was ever undertaken on that site, but for records, the site is not deleted.

With respect to the grammar, I'm a huge fan of the oxford comma, they clear a lot of ambiguity in some of these lists. I'm not sure we need to get rid of them, in my personal opinion. I'm on board with the corrections of spelling mistakes, though.

LizDerrington:
Maybe remove this 'Never Used' discussion from this pull request.
I agree that all misspelt words should be corrected in vocabularies but do not think punctuation corrections are warranted unless the meaning of the sentence is unclear. 
As Andrew has commented above, we all punctuate differently based on preferences. My view is that the sentences as they appear are not ambiguous and therefore do not require remediation.

johnmckellar:
Hi Liz, Andrew,

We need to follow the GSQ Style Guide (attached), in this case for the use and non-use of serial commas. Yes, we all punctuate differently, and can do so correctly. This is why we have such a guide.

I fully support the use of serial commas, where if not used, a qualification of a preceding term/noun can flow on to the succeeding term/noun, thus causing ambiguity, etc. So, I ask you to please follow the guide.

I will deal with the discussion re 'Never Used' separately. It's part of the discussion in the PR, but is not part of a change requested in the PR. I need to attend to another matter first.

Cheers,

John 👍

GSQ-AI:
> What do you think about modifying it to the following, to eliminate slight ambiguity? The status of an individual site (e.g., mine, well) or single activity (e.g., survey) at a site describing the current state of the site or the site activity, within its lifecycle from proposed through operational through to abandonment. Lifecycle stages are commonly related to regulatory conditions administered by government agencies.

We can probably eliminate some redundancy by breaking that first sentence into two, and deleting repeated terms.

Perhaps:

`The status of an individual site (e.g. mine, well) or site based activity (e.g. survey). It describes their current status within their lifecycle (From proposed to operational, through to abandonment). Lifecycle stages are commonly related to regulatory conditions administered by government agencies.`

johnmckellar:
HI Andrew,
I have updated the PR as per your suggestion of breaking the first sentence into two.
I saw your reply when I went into GH this morning and thought that I did not receive a GH alert re it. But, alas, in scanning my emails, I found that this was not the case, highlighting my oversight.
Where to from here re the rest?

johnmckellar:
I have closed the PR and deleted the branch, as, on uploading the file, there appears to have been other changes made since my original PR (or VS has constructed the file differently from what I downloaded to it today, perhaps this being more likely the case). I am not prepared to make sense of this, as the PR has been sitting here too long. Andrew, I'll leave it you to correct the misspelling/s and reconstruct the ambiguous sentence (outlined by me), as per your suggestion, which I agree with.
Cheers,
John

PR #465, opened 2023-03-10
QLD Gov vocabs validated

edmondchuc:
All vocabularies are now passing VocPub v3 rules.

All vocabulary files have been formatted with the [longturtle serialiser in RDFLib](https://rdflib.readthedocs.io/en/stable/plugin_serializers.html). Future edits to the vocabularies should format modified files to ensure clean git diffs.


GSQ-AI (approved):
As with the other pull requests in this series, this seems okay in raw ttl, but will benefit more from testing in UAT.

Steven-Bowden (approved):
This validation work will need to be system tested UAT

PR #474, opened 2023-03-30
LPX+ additions for geofeatures

GSQ-AI:
Added two new vocabularies for LPX+ UAT, as requested by SRA.

geological-confidence.ttl
geological-property-type.ttl

The definitions are placeholder until such time as staff are available to scrutinize them.


LukeHauck (approved):
These vocabs make sense and tech all good.

Steven-Bowden (approved):
looks good, we can build upon this over time.

PR #475, opened 2023-03-31
Lpx plus additions redundant vocab removals

Steven-Bowden:
These two files are not necessary as other vocabs already exist to get the same outcome.


GSQ-AI:
I can confirm that we found that there we two vocabs which perform these same roles;

https://vocabs.uat.gsq.digital/object?uri=http%3A//linked.data.gov.au/def/qualitative-confidence
https://vocabs.uat.gsq.digital/object?uri=http%3A//linked.data.gov.au/def/geological-properties

As such, I approve their deletion.

GSQ-AI (approved):

LukeHauck (approved):
Deletion approved.

PR #481, opened 2023-05-11
Bporehole Status: Add missing labels

nicholascar:


Steven-Bowden (approved):
Looks good, thanks.

GSQ-AI (approved):
The changes make sense to me. I don't see this introducing any additional ambiguity or issues.

I would like to note that on line 94 in the update (Line 99 in the original), that "Drilling" is an alt label for "Drilling Completed" in Western Australia. Given that linkage exists in the un-updated vocabulary, I'm happy to let it be, but it might be worth double checking that relationship is correct. It just feels odd to me that the status of drilling completed has the verb drilling as an alt-label.

PR #484, opened 2023-06-11
QSI CAM vocabs

nicholascar:
Seven road vocabularies and two geographical names vocabularies from Queensland Spatial Information.

These vocabularies will appear to be invalid since they fail validation according to the version of the validator stored in this repo but they are valid when using the latest version of the validator which should be updated from https://w3id.org/profile/vocpub/validator.ttl.

A single typo in Geologic Feature Types is also fixed.

### DIFF
--- a/vocabularies-gsq/geologic-feature-types.ttl
+++ b/vocabularies-gsq/geologic-feature-types.ttl
@@ -1286,14 +1286,14 @@ geof:TectonicPlate
     skos:prefLabel "tectonic plate"@en ;
 .
 
-geof:TectonisedMetomorphosedProvince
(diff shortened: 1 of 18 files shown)


Steven-Bowden (changes requested):
Getting validation errors on the QSI vocabs, @nicholascar @edmondchuc can you advise on next action.

edmondchuc:
I've created a PR https://github.com/geological-survey-of-queensland/vocabularies/pull/491 which fixes the issues here. If you merge that PR, it will merge the changes into this PR. The test in this PR should then pass and you can do a final review before merging.

PR #488, opened 2023-06-22
Update borehole-status definition

johnmckellar:
Liz, I have altered the definition for the vocab per your suggestion.


Steven-Bowden (approved):
Approved

LizDerrington (approved):

PR #491, opened 2023-07-21
Fixes validation issues for PR #484

edmondchuc:
This PR fixes validation failures for PR https://github.com/geological-survey-of-queensland/vocabularies/pull/484.

It also updates VocPub to the latest version, which is currently 3.1.

Note that a bunch of files needed to be updated to pass the 3.1 validator. It was primarily a change from using `dcterms:provenance` to `skos:historyNote`.


Steven-Bowden (approved):
looks good

LukeHauck (approved):
Same change for every concept throughout. Looks good. Approved.

PR #502, opened 2023-09-18
Update geounits.ttl for cm topConcept

johnmckellar:
@robchatterjee-resources @LizDerrington

Hi Nick, (Andrew),

FYI:
@Liz Derrington Received the following message from SRA.

After reviewing the vocabs it appears to be the issue with the ttl file which we use to upload to PREZ (These files are handled by DOR and need to review it).

The format in the .ttl file seems little different compared to other .ttl file to get the concept

Vocabs (cubic centimeter, Centimeter, Centimetre, Cubic Meter per Hour, Kilotonnes, Liter, Meter, Millimeter, Million Barrels) can bee seen in the .ttl file that is uploaded in GitHub.

Also, the reason why there are two Centimeter, Centimetre is:

These vocabs are derived from https://qudt.org/vocab/unit/CentiM and there are two labels for this vocab one is Centimeter(en-us) and another is Centimetre(en) (Please refer to the below image) which is why they are showing it twice. 

DOR might need to review this .ttl file from GitHub as skos:hasTopConcept section is missing these concepts which are being used to show it in  http://linked.data.gov.au/def/geou

### DIFF
--- a/vocabularies-gsq/geounits.ttl
+++ b/vocabularies-gsq/geounits.ttl
@@ -17,7 +17,8 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
     skos:hiddenLabel "CENTIMETRES (cm)"@en ;
     skos:inScheme cs: ;
     skos:notation "cm" ;
-    skos:prefLabel "Centimeter"@en ;
+    skos:prefLabel "Centimetre"@en ;
+    skos:topConceptOf cs: ;
 .
 
 geou:MegaBBL


johnmckellar:
Andrew, Nick, @robchatterjee-resources 
The problem is with:
http://qudt.org/vocab/unit/CentiM
which gives the rdf labels for centimeter en-us & centimetre en

The solution here would be to change the concept from
http://qudt.org/vocab/unit/CentiM
to
geou:CentiM
and add this to the list of concepts at the bottom of the file. Would this be the solution or is there more to this than this fix?

GSQ-AI:
Unless @nicholascar has a solution for the QUDT bringing back both values (and not displaying at all in Prez), I think that you're right John, and that changing the vocab so that we don't pull the concepts from QUDT is the best path to a resolution.

nicholascar:
John's proposed change was correct but not sufficient: we also need `<ConceptScheme> skos:hasTopConcept <TheConcept>` in addition to his proposed `<Concept> skos:topConceptOf <ConceptScheme>`. Yes, ontologically there's redundancy there but it's easier to mandate lots of properties in the validator than to perform reasoning.

I've addressed this change with a couple of direct commits to the master branch so that I could also fix a validator error. All done now in Commits https://github.com/geological-survey-of-queensland/vocabularies/commit/4e360e40b4f0f526e58dd1bf83fd6494158d7954 & https://github.com/geological-survey-of-queensland/vocabularies/commit/8b370820b0e6527ca3ee825c5546a563007ad010 so this PR may be closed without merging now.

johnmckellar:
Thanks Nick

PR #504, opened 2023-09-27
Update sample-type.ttl

geoderekh:
removed hyperspectral scan as this is better modelled as an observation type on a core or cuttings sample type. Happy to present my rationale and evidence of the approved business procedure to a group if needed.

### DIFF
--- a/vocabularies-gsq/sample-type.ttl
+++ b/vocabularies-gsq/sample-type.ttl
@@ -294,17 +294,6 @@ sampty:hand-sample
     skos:topConceptOf cs: ;
 .
 
-sampty:hyperspectral-scan
-    a skos:Concept ;
-    skos:historyNote "Compiled by the Geological Survey of Queensland" ;
-    rdfs:isDefinedBy cs: ;
-    skos:definition "A composite image comprised of multiple wavelength bands across the electromagnetic spectrum."@en ;
-    skos:inScheme cs: ;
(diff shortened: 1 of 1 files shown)


Greenwoodmatthew (approved):
I agree that the hyperspectral scan is an observation type and not a sample type.  For Hyperspectral i believe the model is that the core is a sample of the borehole, and the scan is a observation of the core

LukeHauck (approved):
Tech good. Approved.

Greenwoodmatthew:
Hi all. Checking on status of this.  Ready to merge? @Steven-Bowden do you need to review?

geoderekh:
@Steven-Bowden  I notice that geoproperties PROD still recognises Hyperspectral Scan as a valid sample type. This issue seems to pop up regularly, so can we align the vocab PR-commit-merge procedure with the following steps to update the graph DB and SRA service request to refresh the Geoprops server. I thought that this had been/was going to be automated by Nick.

Greenwoodmatthew:
@Steven-Bowden is there a final step that id missing between committing the cose to mainin github and picking up the new vocabs in prod?  Is this an SRA task, or a GSQ task?

Steven-Bowden:
GSQ task is being resolved as per GSD-480

Steven-Bowden:
resolved https://vocabs.gsq.digital/v/vocab/def:sample-type

PR #515, opened 2023-12-01
update timescale to use official ICS release

nicholascar:
Replace the previous test SKOS version of the Geologicla Timescale with the Chronostrat Chart official release from the ICS. Time periods use the same IRIs


robchatterjee-resources (approved):
Approve

GSQ-AI (approved):

johnmckellar:
@nicholascar @GSQ-AI @robchatterjee-resources @Steven-Bowden 

Happy New Year everyone!!

I saw this PR when I was on leave and thought that it would not be dealt with until after I returned (today).

This ontology needs a little reworking as it seriously mixes chronostratigraphic and geochronologic units. This is easy to do, given the way the International Chronostratigraphic Chart (Geological Time Scale/GTS) is presented by the ICS/IUGS. It is a universal problem and one that the ICS needs to address ... and it can be done with only slight modification to the GTS chart.

Both types of units are shown in the chart's column headings, but the principal problem is for the 'Series/Epoch' column where both types of units are not shown in the column itself, only the chronostratigraphic unit, not also the geochronologic one.  By this I mean that 'Upper' should be 'Upper/Late' and 'Lower' should be 'Lower/Early' (to conform with the heading 'Series/Epoch'). By these omissions, the chart implies that the terms 'Upper' and 'Lower' apply to both Series and their Equivalent Epochs, this being the source of ongoing international confusion (and one of the problems in our vocab).

Also, the ICS should call the chart the 'International Chronostratigraphic and Geochronologic Chart', as both sets of units are encompassed by the chart, although the geochronologic units are based on the chronostratigraphic units (e.g., for the Phanerozoic, Ages are based on Stages, the base of each of the latter being ultimately defined (where possible) by their Global Boundary Stratotype Section and Point/GSSP).

These suggestions would then cause users to stop and think about what the terms mean. Better still, rather than this mouthful for the chart's name, the ICS should refer to the online chart as the 'Geologic Time Scale' or the 'International Geologic Time Scale', the former being what the ICS called the chart in the major printed version, which was published every eight years, and in the intervening minor update/version, which was printed every other four years. To my knowledge, the last major printed version was:  ‘A Geologic Time Scale 2012’ by Gradstein et al. (2012). This, by the way, is in the caption to the present online version of the chart: https://stratigraphy.org/ICSchart/ChronostratChart2023-09.pdf

Nick, you are in a position with the ICS to convince them to do something about all of this.

Before I attempt any edits to the file to correct the cited and other problems, please let me know what you think ... and I will then proceed.

Cheers

PR #517, opened 2023-12-05
Remove Extractives Statistical reporting from visibility in LP - Update georesources-report-types.ttl

Steven-Bowden:
Extractives Statistics should not be visible on public and internal LP


GSQ-AI (approved):
This seems sound, good to deploy based on tests in UAT.

robchatterjee-resources (approved):

LizDerrington:
Approved

PR #521, opened 2023-12-20
Add GSQ details to fix Borehole Status validity

nicholascar:


nicholascar:
Merging without review as this is just making this vocab valid where before it was missing agent inf (GSQ)

PR #522, opened 2024-01-04
Revised geological-observation-type.ttl

geoderekh:
fix duplicated positional observation concept definition per issue #456 and correspondence with AI and MG

PR #523, opened 2024-01-04

Steven-Bowden:
Updated definition of positional as per PR #522 , using recent branch to streamline merge.


GSQ-AI (approved):
Sensible change to update from the generic observation definition, which doesn't apply to this topic.

Approved.

Greenwoodmatthew (approved):

PR #525, opened 2024-01-11
Update survey-relationship-type.ttl

LukeHauck:
In GeoProps, we can only link seismic surveys one-way with "Reprocessing" i.e. In the case where Survey 3 is a reprocessing of Survey 1 and Survey 2, the related survey records in GeoProps for Surveys 1 and 2 would both indicate that Survey 3 is a reprocessing.

However, there is no relevant option to update Survey 3 to indicate that 1 and 2 are the original surveys. 

I suggest adding in "Reprocessed" so we can clearly articulate the directional link i.e. Survey 3 would be updated to indicate that Surveys 1 and 2 are related reprocessed surveys.

### DIFF
--- a/vocabularies-gsq/survey-relationship-type.ttl
+++ b/vocabularies-gsq/survey-relationship-type.ttl
@@ -59,21 +59,21 @@ srt:repeat
     skos:prefLabel "Repeat"@en ;
 .
 
-srt:reprocessed
+srt:reprocessing
     a skos:Concept ;
     skos:historyNote "Developed by the Geological Survey of Queensland." ;
     rdfs:isDefinedBy cs: ;
-    skos:definition "The original entity from which a new digital sample was created by reprocessing."@en ;
+    skos:definition "The relationship type applied to a related reprocessed survey to indicate that survey is the reprocessing job. Only used for related surveys within an original survey record to identify a related reprocessed survey."@en ;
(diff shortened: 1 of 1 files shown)


GSQ-AI (dismissed):
Happy for this to go forward as is, but potential changes that might be worth considering are:

Making this or reprocessing a narrower concept ie.

Reprocessing
--Reprocessed

`skos:broader srt:reprocessing `

The other suggesting being something I'm not completely sold on being valid, but it could potentially make the the concept relationship clearer; append 'as' to the concept title, becoming "reprocessed as".

But, as mentioned before, I'm happy to go ahead without any modification. Just wanted to offer some thoughts on potential improvements.

Steven-Bowden (approved):
Looks fine, a good candidate for bidirectional relationship.

LukeHauck:
After discussion with Andrew, it was decided to more clearly delineate the two concepts.

Within an original survey record, I add a related reprocessed survey. So I add the relationship type "reprocessing" to indicate this related survey is a reprocessing of the original.

Within a reprocessed survey record, I add a related original survey. So I add the relationship type "reprocessed-from" to indicate this related survey is an original that was reprocessed.

Steven-Bowden (approved):
looks good

GSQ-AI (approved):

PR #528, opened 2024-01-22
Update geological-observation-method.ttl

LukeHauck:


Steven-Bowden (changes requested):
on reflection, the collections shouldn't be in this hasTopConcept list

LukeHauck:
Collections removed from list as suggested

Steven-Bowden (approved):
looks good

GSQ-AI (approved):

PR #533, opened 2024-02-06
Update geological-observation-method.ttl

Steven-Bowden:
when reviewing in UAT, noticed a few top concepts were missing from list.


DevenderKumarGSQ (approved):
Passed all checks.

PR #535, opened 2024-02-26
Standardising Terms Across Systems

GSQ-AI:
Two small changes;

"Capped and Abandoned" becomes "Plugged and Abandoned" in order to coincide better with GeoProperties. The AltLabel says both are the same, but ensuring that the names are the same in both definition and system limits confusion.

"On-Production" becomes "Producing" to align the definition with other vocabularies and the terms used in GeoProperties.


LukeHauck (approved):
Changes make sense and the syntax is good. Approved.

Steven-Bowden (approved):
Looks good

PR #539, opened 2024-03-13
Implement Key Resource Area Report

GSQ-AI:
Changing report types to include "key-resource-area-reports", change made in conjunction with changes to reportTemplates.json:
https://github.com/geological-survey-of-queensland/gsq-lodgement-portal/blob/addingKRAreports/scripts/reportTemplates.json

Key Resource Area Reports sit in the non-permit reports and department-reports vocabulary collections, for internal display only and to separate from the required spatial components. Could be updated in the future to make .shp files describing boundaries of each KRA mandatory.

### DIFF
--- a/vocabularies-gsq/georesources-report-types.ttl
+++ b/vocabularies-gsq/georesources-report-types.ttl
@@ -49,6 +49,7 @@ grt:departmental-reports
         grt:geological-survey-of-queensland-publication ,
         grt:geological-survey-of-queensland-record ,
         grt:industry-consultative-report ,
+        grt:key-resource-area-report ,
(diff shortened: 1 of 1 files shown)


Steven-Bowden (commented):
I notice this is not in grt:lodgement-portal-internal ( a skos:Collection ) are we looking to have this lodged through the front-end?

LukeHauck (dismissed):
Changes make sense and vocab is coherent.

Steven-Bowden (approved):
looks good

DevenderKumarGSQ (approved):
Looks good

PR #545, opened 2024-04-04
Fix missing top Concepts

nicholascar:
Rock Unit Rank is missing top concept declarations for most Concepts so they are not displaying, see;

https://vocabs.gsq.digital/v/vocab/defn:rock-unit-rank

This PR just adds them all to the top of the vocab, so all will be listed.

further hierarchy can be added once they are visible, if needed.


GSQ-AI (approved):
The diff has gone a bit haywire, but, to my understanding, this change looks good.
Approved.

Steven-Bowden (approved):
small addition of code, looks fine.

PR #553, opened 2024-06-20
Pet stats vocab

Steven-Bowden:
add survey type for reserves


LukeHauck (approved):
Tech good, makes senses, approved.

GSQ-AI (approved):
Approved, no alternative or more elegant solutions are available, and it aligns with existing survey types.

PR #555, opened 2024-06-26
Addr model updates

nicholascar:
Updates for the Address Model:

* Address Part Type vocab - new
* Updated custodian info for all vocabs - using Data Roles vocab, not ISO to indicate role

Address Part Types will need further updates - this is just a start.

### DIFF
--- a/validation/validate_all.py
+++ b/validation/validate_all.py
@@ -20,7 +20,7 @@ import json
 import sys
 
 # set this to the vocab folder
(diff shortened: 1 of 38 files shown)


GSQ-AI (approved):
Approved.

Steven-Bowden (approved):
Looks fine technically, QSI changes outside of my subject domain

PR #556, opened 2024-08-01
QSI vocabs review 2024-08-01

annegoldsack:
Review of QSI vocabs with some updated to definitions plus removal of some specific, and creation of some more generic vocabs.


nicholascar (approved):

Steven-Bowden (approved):
Looks fine

PR #559, opened 2024-08-06
WIMS experimental vocab

GSQ-AI:
WIMS experimental vocab.


Steven-Bowden (approved):
Great start, approved as experimental.

LukeHauck (approved):
Makes sense, and workable. Approved as experimental vocab.

PR #562, opened 2024-08-13
Fix QG Organisations

broughd-qld:
Slight error in file that didn't align units with new org.


GitVillain (approved):
good to go

Steven-Bowden (approved):
Make sense, approved.

PR #563, opened 2024-08-13
Update retention-status.ttl

GSQ-AI:
WIMS - Fix notation


LukeHauck (approved):
Make sense. Approved.

Steven-Bowden (approved):
Thanks

PR #572, opened 2025-04-15
Remove all QSI vocabs

nicholascar:
Qll Qld Spatial Information and ICSM vocabs can now be removed since they are online at https://dev.icsm.information.qld.gov.au/ and will soon be in production at https://icsm.information.qld.gov.au/.

The listing of "QSI vocabs" at https://vocabs.gsq.digital/v should also be removed.


LukeHauck (approved):
Looks good. Approved.

PR #575, opened 2025-05-30
Access rights update 2025 05

johnmckellar:
Update of Data Access Rights for samples for WIMS, as requested.


johnmckellar:
@Steven-Bowden @geoderekh For the two collections, my I suggest that there is a scopeNote providing the reasons for the collections.

geoderekh (approved):
Changes look good. I like the use of scopenote to elucidate the application of the vocab to samples.

PR #579, opened 2025-06-24
Update georesources-report-types.ttl

LukeHauck:
To prepare for a soft launch, based on feedback and advice.

Removed reports from non-current.

Added reports into visible to external users.


Steven-Bowden (approved):
makes sense based on new approach.

GSQ-AI (approved):
Looks good, approved.

PR #581, opened 2025-07-09
Update gsq-alias.ttl

geoderekh:
A quick turnaround is preferred as this PR blocks the completion of the SQL view and TDI data migration. 
Scope: Added three aliases (WIMS-sample-id, GSQR number, GSQM number) to enable the WIMS and WIMS TDI projects.


vocabularies-gsq/gsq-alias.ttl
```diff
@@ -44,6 +44,7 @@ als:government
         als:mmol ,
         als:ogia ,
         als:treasury ;
+        als:wims-sample-id ;
```
GSQ-AI, line 47: Why should wims-sample-id be included in this collection?  I may be wrong, but it doesn't appear consistent with the other concepts. A 'WIMS' concept would fit here, in my opinion, but might be premature before the system is named officially.

vocabularies-gsq/gsq-alias.ttl
```diff
@@ -311,6 +344,7 @@ cs:
         als:rshq ,
         als:service-provider ,
         als:treasury ,
+        als:wims-sample-id ,
```
GSQ-AI, line 347: GSQ R and M number concepts should be included here as well, shouldn't they?

LukeHauck (dismissed):
Change makes sense. Approved.

vocabularies-gsq/gsq-alias.ttl
geoderekh: There is a GSQ legacy concept, which doesn't appear here, so these appear to be other government organisations. GSQR and GSQM are narrower concepts of GSQW legacy.

vocabularies-gsq/gsq-alias.ttl
geoderekh: WIMS is akin to MMOL, which is a system from another business group, hence it's inclusion here. 
We must deliver the SQL view by the 15th, and this is the final blocker for me to finalise the SQL. I don't think there is any chance WIMS will be renamed on completion

geoderekh: Any inconsistency here is likely due to a lack of specificity in the other concepts. 
Take MMOL for example, there are many identifiers in that database, so MMOL as an alias source is useless. MMOL permit id, or mmol subblock id might be more relevant. 
WIMS sample id is the alias in the wims system, and calling this alias WIMS would make it unusable in geoprops since we can only specify and alias value.

geoderekh:
@Steven-Bowden Who manages the validation? Test are throwing a bad syntax error but I can't see where it is coming from and the ttl file looks good to me.

vocabularies-gsq/gsq-alias.ttl
```diff
@@ -44,6 +44,7 @@ als:government
         als:mmol ,
         als:ogia ,
         als:treasury ;
```
GSQ-AI, line 46: ; needs to be a comma

geoderekh: fixed.

GSQ-AI (approved):
Happy with the state the vocab is in now and with the considerations made.
Approved.

LukeHauck (approved):

PR #586, opened 2025-09-16
Update sample-type.ttl - add core run concept

GSQ-AI:
Introduce the 'core-run' concept as a narrower concept of the core, describing the discrete samples that make up the core core concept. Core-run concept is experimental, but necessary to assign the existing core runs to in the GeoProperties system as a part of the WIMS data remediation project.


Steven-Bowden (approved):
looks good, will be useful

PR #587, opened 2025-09-18
Update sampling-method.ttl

GSQ-AI:
Revert https://pid.geoscience.gov.au to http//pid.geoscience.gov.au to support GDMP system validation.


Steven-Bowden (approved):
looks good, consistent with system and target vocab e.g https://vocabs.ardc.edu.au/repository/api/lda/ga/sampling-method/v1-0/resource?uri=http://pid.geoscience.gov.au/def/voc/ga/samplingmethod/core_drilling_unspecified

kalaldisha (approved):
Approved

PR #593, opened 2026-03-03
fix IRIs for rdflicense

nicholascar:
The IRIs for the linked-to RDF Licenses were incorrect. Here fixed.


Steven-Bowden:
@nicholascar Hi Nick, is the exact_match URI meant to change as well? Also, is there a reference for this update, to support the change?

PR #596, opened 2026-05-12
Add elevation observation type to geological vocabulary

Greenwoodmatthew:
Added 'elevation' concept with definitions and labels.

Required to allow LIDAR surveys to be in geoprops --> CKAN in future

### DIFF
--- a/vocabularies-gsq/geological-observation-type.ttl
+++ b/vocabularies-gsq/geological-observation-type.ttl
@@ -150,7 +150,7 @@ obstp:survey-obs
         obstp:directional ,
         obstp:electrical ,
         obstp:electromagnetic ,
-        obstp:elevation
+        obstp:elevation ,
         obstp:geochemical-signature ,
         obstp:geothermics ,
         obstp:gravity ,
@@ -963,7 +963,7 @@ cs:
         obstp:drilling-int ,
         obstp:electrical ,
         obstp:electromagnetic ,
-        obstp:elevation
+        obstp:elevation ,
         obstp:fabric ,
         obstp:fossils ,
         obstp:geochemical-signature ,
--- a/vocabularies-gsq/geological-observation-type.ttl
+++ b/vocabularies-gsq/geological-observation-type.ttl
@@ -150,6 +150,7 @@ obstp:survey-obs
         obstp:directional ,
         obstp:electrical ,
         obstp:electromagnetic ,
+        obstp:elevation
         obstp:geochemical-signature ,
         obstp:geothermics ,
         obstp:gravity ,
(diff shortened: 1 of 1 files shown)


Steven-Bowden (dismissed):
looks good

CantRoger (dismissed):
Looks good. Good addition to the observation types for surveys.

vocabularies-gsq/geological-observation-type.ttl
```diff
@@ -921,6 +922,23 @@ obstp:photographic
...
+obstp:elevation
+    a skos:Concept ;
+    skos:historyNote "Compiled by the Geological Survey of Queensland" ;
+    rdfs:isDefinedBy cs: ;
+    skos:altLabel
+        "Altitude Observation"@en ,
+        "Height Measurement"@en ,
+        "Vertical Position"@en ,
+        "Elevation Survey"@en ,
+        "Topographic Height"@en ;
```
GSQ-AI:
```suggestion
```
Suggest we remove topographic height from altlabel list; 
Topographic Height isn't quite the same as Elevation and for simplicity we can just leave it out.

CantRoger: Perhaps we could replace with more specific term, such as Orthometric Height, as opposed to ellipsoidal or geoidal heights which all fall under the broad 'topographic height'.

GSQ-AI: Orthometric would be a fine addition/edit:
> The orthometric height (symbol H) is the [vertical distance](https://en.wikipedia.org/wiki/Vertical_distance) along the [plumb line](https://en.wikipedia.org/wiki/Plumb_line) from a point of interest to a reference surface known as the [geoid](https://en.wikipedia.org/wiki/Geoid), the [vertical datum](https://en.wikipedia.org/wiki/Vertical_datum) that approximates [mean sea level](https://en.wikipedia.org/wiki/Mean_sea_level).[[1]](https://en.wikipedia.org/wiki/Orthometric_height#cite_note-1)[[2]](https://en.wikipedia.org/wiki/Orthometric_height#cite_note-2) Orthometric height is one of the scientific formalizations of a layman's "[height above sea level](https://en.wikipedia.org/wiki/Height_above_sea_level)", along with other types of [heights in Geodesy](https://en.wikipedia.org/wiki/Geodesy#Heights).

Greenwoodmatthew: This sounds like a good addition.  I will update PR now

Greenwoodmatthew (commented):
Updated with Orthometric height

CantRoger (approved):

GSQ-AI (approved):
Looks good, approved!

geoderekh (approved):
Modelling looks good to me.

PR #599, opened 2026-06-18
Update ANZIC2006-industry-classifications.ttl

GSQ-AI:
Updated to 'coal' as per INTEL request. 
This will cause issue if this entry is to be used for any other commodity in the future, and is in part, contrary to the definition.


Greenwoodmatthew (approved):
minor change, no concerns

Steven-Bowden (approved):
Temporary update approved, should be an alias. Awaiting SPARQL lookup wiring to be updated before restoring generic prefLabel.

PR #600, opened 2026-09-23
Update geo-commodities.ttl

GSQ-AI:
The most basic implementation of hydrogen into the geo-commodities vocab.

Further updates could include the definitions of green, brown and whatever hydrogen, liquid and gaseous hydrogen or a more detailed description of the element.

I'm not sure it needs to be in the lodgement collection, given the behaviour of 5500NAR, but for safety's sake I've included it.


Steven-Bowden (approved):
Suitable for initial establishment

LukeHauck (approved):
Approved.

